Issue #3, opened 2022-09-12
Make the API generic - removing WIC references


chouinar:
After https://github.com/navapbc/template-application-flask/issues/2 - go through and remove the WIC MT specific naming and implementation including:

* Documentation
* API implementation
* Script implementation

Note this should make an equivalent, but not fully formed implementation (eg. missing thorough examples).

PR #8, opened 2022-09-16
First commit containing the base template for a Flask API

chouinar:
https://github.com/navapbc/template-application-flask/issues/3

Reviewed in https://github.com/navapbc/wic-mt-demo-project-mock-api/pull/20

Just copying over the generic version from the above PR. This contains:
* API with a post endpoint that writes to the DB using basic apikey authentication.
* Postgres DB + alembic migrations configured for updating the DB when you update the sqlalchemy DB models.
* Thorough formatting & linting tools.
* Logging, with formatting in both human-readable and JSON formats.
* Ability to write backend scripts connected to the DB with an example one that generates a CSV (locally or on S3 with proper credentials).
* Ability to run the various utility scripts inside or outside of Docker (Note: I may need someone to test the outside as it requires installing a few libraries and/or configuring a few things that I need to document).
* Restructured and improved API request and response error handling which gives more details than the out-of-the-box approach for both Connexion and Pydantic
* Sane environment variable configuration for local development (`local.env`)


app/api/route/response.py
```diff
@@ -0,0 +1,81 @@
...
+
+import flask
+from werkzeug.exceptions import HTTPException
+
+
+@dataclass
+class ValidationErrorDetail:
+    type: str  # In the Mass code this is an enum, but lets keep it basic here
+    message: str = ""
+    rule: Optional[str] = None  # Also an enum in Mass
```
rocketnova: Let's drop these comments

rocketnova (approved):
One comment. Otherwise LGTM!

Issue #4, opened 2022-09-12
Add additional example API endpoint types


chouinar:
Once https://github.com/navapbc/template-application-flask/issues/3 has been done, go through and add additional API examples for other non-POST endpoints.

PR #10, opened 2022-09-20
Adjust the example implementation to be user+role

chouinar:
## Ticket

https://github.com/navapbc/template-application-flask/issues/4

## Changes
Adjusting the example API implementation to use User+Role instead of ExamplePerson+ExamplePet to be a more realistic baseline.

## Context for reviewers
The existing examples weren't quite something a developer could use as a baseline, so changed them to be User+role rather than Person+Pet. The actual shape of the models is pretty similar except for adding in the Lookup Table approach to handle the enums. tl;dr; SQLAlchemy + Alembic + Enums are really clunky - Alembic can't detect changes due to how SQLAlchemy implemented enums + Postgres enums have a few caveats. See some details in the comments of this PR for more details.

The implementation should be basically the same as before otherwise, just with a slightly different set of models.

As no one is directly using this, I deleted the original migration to entirely remove the Person/Pet references.

## Testing
I went through each of the API, tests, and CSV generation to verify that everything worked as before.


app/api/db/models/lookup.py
```diff
@@ -0,0 +1,186 @@
...
+#
+
+from typing import Any, Dict
+
+import api.logging
+
+logger = api.logging.get_logger(__name__)
+
+
+class LookupTable:
```
chouinar: This is 100% verbatim from the Mass approach. I tried to think of another approach to creating a static list of values, but kept coming back around to this approach. Open to suggestions if there's another common pattern. I do like that the static values live in the DB which makes direct SQL queries a bit more straightforward from past experience.

app/api/db/migrations/versions/2022_09_20_13_16_20_dbc5b91646a2_create_user_table.py
```diff
@@ -0,0 +1,96 @@
...
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.Column(
+            "updated_at",
+            sa.TIMESTAMP(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.Column("user_id", postgresql.UUID(as_uuid=True), nullable=False),
```
lorenyu: what are your thoughts on the naming convention of the id being `id` rather than `[table]_id`? it seems redundant to repeat the table name again.

app/api/db/models/lookup.py
lorenyu: what are the benefits of having a lookup tables rather than an enum type in SQL?

app/api/db/__init__.py
```diff
@@ -12,12 +12,15 @@
...
 from api.db.migrations.run import have_all_migrations_run
+from api.db.models import init_lookup_tables
 
 logger = api.logging.get_logger(__name__)
 
 
 def init(
-    config: Optional[DbConfig] = None, check_migrations_current: bool = False
+    config: Optional[DbConfig] = None,
+    sync_lookups: bool = False,
```
lorenyu: nit: rename to `sync_lookup_tables` but let's first resolve the discussion about whether to use lookup tables in the first place

lorenyu (commented):
left a comment to discuss the value if lookup tables as compared to the simpler enum type

app/api/db/models/lookup.py
chouinar: I was not aware of the enum type in SQL (why didn't we use that in Mass?) - just sorta assumed it didn't exist and never checked. Let me see how that works.

chouinar: Tinkering with SQLAlchemy/alembic/postgres enums. They work pretty well - define python enum class, automatically gets created in the pg_enum table, can be used by multiple different tables which will all reference the same enum in the DB. Feels pretty intuitive and straightforward to work with

The primary issue is altering an existing enum. If I change the enum python class, alembic doesn’t auto-detect the changes to alter the enum in the DB, and every guide I lookup online just tells you how to write the SQL to do the alter command manually - defeating the purpose of alembic. Might be able to build that auto-detect manually into the alembic scripts.

Consequently, enum values can’t be removed, so a down migration can’t remove an existing enum value, although technically our lookup tables worked in the same way. You can still entirely delete an enum set though.

chouinar: Looks like you can alter enum values in postgres as well: https://blog.yo1.dog/updating-enum-values-in-postgresql-the-safe-and-easy-way/ but not sure if it's worth the hassle.

@lorenyu  Any thoughts on how we should progress here? That not recognizing an alteration problem would need some solution to make this feel like a good approach. Should we stick with the lookup table approach, or dive further into enums (might be out of scope for the time I have to get this done)?

chouinar: I did find someone who figured out auto-migrating for enums: https://github.com/dw/alembic-autogenerate-enums/blob/master/alembic_autogenerate_enums.py

chouinar: I'm going to push forward as-is, but I did put together a quick branch which shows enums with the above library. The down migration is misleading however as it does nothing.
https://github.com/navapbc/template-application-flask/blob/chouinar/enum-example/app/api/db/models/user_models.py

Creates the following migration: https://github.com/navapbc/template-application-flask/blob/chouinar/enum-example/app/api/db/migrations/versions/2022_09_21_10_44_26_f9e1862f9c52_create_user_tableagain.py and with the adjusted migration script, and an adjustment to the enum produces: https://github.com/navapbc/template-application-flask/blob/chouinar/enum-example/app/api/db/migrations/versions/2022_09_21_13_21_15_389e7f257581_add_another_enum_val.py

lorenyu: @chouinar sounds good to proceed as-is for now.

re: the other branch, do we still need "sync_lookup_tables" with enums?

chouinar: No, wouldn't need that, I just switched to the bare minimum to get it working, but all the lookup table specific logic isn't necessary.

app/api/db/migrations/versions/2022_09_20_13_16_20_dbc5b91646a2_create_user_table.py
chouinar: I think it can help when writing queries that join across several tables to know which ID you're working with / make you less likely to make a mistake for joins. Which of these would you find more readable?

```sql
SELECT
	*
FROM
	person
	JOIN place ON place.place_id = person.place_id
	JOIN thing ON thing.thing_id = place.thing_id
```
or if each table has just ID
```sql
SELECT
	*
FROM
	person
	JOIN place ON place.id = person.place_id
	JOIN thing ON thing.id = place.thing_id
```
(I actually had to retype that a few times because I kept doing `JOIN place ON place.place_id = person.id` - but that could just be mad muscle-memory)

app/api/route/user_route.py
```diff
@@ -0,0 +1,42 @@
...
+    return response_util.success_response(
+        message="Success", data={}, status_code=200
+    ).to_api_response()
+
+
+# Get user
+def user_get(user_id: str) -> flask.Response:
+    return response_util.success_response(
+        message="Success", data={}, status_code=200
+    ).to_api_response()
```
chouinar, line 42: In a follow-up, I'll implement these two endpoints. Was going to do it in this PR, but wanted to keep this contained to purely switching the models used without any meaningful implementation change.

app/api/db/migrations/versions/2022_09_20_13_16_20_dbc5b91646a2_create_user_table.py
lorenyu: the second seems more readable to me, `place.id = x.place_id` since it also reads as "place id equals table place id", vs. "place dot place id equals table place id", there's extra words that don't give me extra info.

app/api/db/models/user_models.py
```diff
@@ -0,0 +1,75 @@
...
+        self.role_description = role_description
+
+
+class Role(LookupTable):
+    model = LkRole
+    column_names = ("role_id", "role_description")
+
+    USER = LkRole(1, "User")
+    ADMIN = LkRole(2, "Admin")
+    THIRD_PARTY = LkRole(3, "Third Party")
```
lorenyu: i don't understand why there's a need for both a `Role` and a `LkRole` type. could you help me understand that?

app/api/db/models/lookup.py
lorenyu: Bringing in the convo from Slack, I think we should keep it simple and get rid of all this boilerplate code and just have Role be a TEXT column and in the application code it could be a class perhaps an [Enum](https://docs.python.org/3/library/enum.html) type in Python (looks like pydantic also just leverages the existing python enum as well)

lorenyu: If you want to do that in a separate PR lemme know I could approve this one for now

chouinar: I went ahead and switched to having the validation purely in the API/Pydantic layer. The DB just has it as a text column.

Note that I did also look into the check constraint for validating the text field, but alembic also can't auto-detect changes to it, so it ends up being effectively the same a the enum field.

app/api/db/migrations/versions/2022_09_22_11_13_20_964e99759832_create_user_and_role_tables.py
```diff
@@ -55,25 +55,23 @@ def upgrade():
...
-            ["example_person.example_person_id"],
+            ["user_id"],
+            ["user.user_id"],
         ),
-        sa.PrimaryKeyConstraint("example_pet_id"),
+        sa.PrimaryKeyConstraint("user_id", "role_description"),
     )
     op.create_index(
-        op.f("ix_example_pet_pet_owner_id"), "example_pet", ["pet_owner_id"], unique=False
+        op.f("ix_user_role_role_description"), "user_role", ["role_description"], unique=False
```
lorenyu: for constraint naming conventions, can we use suffix rather than prefix and `idx` instead of `ix` (e.g. [as described in this stackoverflow answer](https://stackoverflow.com/questions/4107915/postgresql-default-constraint-names/4108266#4108266)
```suggestion
        op.f("user_role_role_description_idx"), "user_role", ["role_description"], unique=False
```

app/api/route/handler/user_handler.py
```diff
@@ -0,0 +1,69 @@
...
+from api.route.api_context import ApiContext
+from api.route.request import BaseRequestModel
+
+logger = api.logging.get_logger(__name__)
+
+
+class RoleEnum(str, enum.Enum):
+    USER = "User"
+    ADMIN = "Admin"
+    THIRD_PARTY = "Third Party"
```
lorenyu: `RoleEnum` is part of the business domain so it may need to be used by business logic classes, so it probably shouldn't live in a file that is tied to flask route handlers (since we don't want business classes depending on things like flask). Could we move this to `user_models.py`?

app/api/db/migrations/versions/2022_09_22_11_13_20_964e99759832_create_user_and_role_tables.py
```diff
@@ -1,16 +1,16 @@
-"""first migration
+"""Create user and role tables first migration
```
lorenyu: For curiosity, what would go in `migrations/` and not in `migrations/versions`, should we flatten the folder structure to just `migrations`?

lorenyu (approved):
Nice!

app/api/db/migrations/versions/2022_09_22_11_13_20_964e99759832_create_user_and_role_tables.py
chouinar: Let me see if there's a way to configure that - this is just the auto-generated format.

lorenyu: Oh if it's autogenerated then nvm that's fine

chouinar: Looks like it was pretty simple to adjust it, just need to adjust a metadata value in the base table class.

app/api/db/migrations/versions/2022_09_22_11_13_20_964e99759832_create_user_and_role_tables.py
chouinar: The migrations folder has alembic configuration and scripts. This is just following the standard structure as generated by it: https://alembic.sqlalchemy.org/en/latest/tutorial.html

PR #15, opened 2022-09-22
Add patch and get endpoint for users

### LINKED ISSUE #4: Add additional example API endpoint types

chouinar:
## Ticket

https://github.com/navapbc/template-application-flask/issues/4

## Changes
This PR adds a GET and PATCH endpoint for the user model to the template repo. This follows up work that added a POST endpoint and related components.

## Context for reviewers
GET endpoint is pretty routine, give an ID, get the model.

PATCH is mostly routine, although the conventions there are slightly less defined. Any unset fields aren't modified, only the set fields. The primary complexity was adjusting roles. It might make more sense to have these as their own separate dedicated endpoint (something like `user-add-role` and `user-remove-role`) if this is deemed too complex. It basically does a set-diff to see what needs to be added/removed to get to the desired state.

Are there any other common-ish endpoint types you would like to see added?

## Testing
Create a user via the POST endpoint:
Screen Shot 2022-09-22 at 4 54 13 PM

You can GET them:
Screen Shot 2022-09-22 at 4 54 40 PM

PATCH request:
Screen Shot 2022-09-22 at 4 55 08 PM

PATCH response:
Screen Shot 2022-09-22 at 4 55 17 PM

GET again after the PATCH:
Screen Shot 2022-09-22 at 4 56 46 PM

### DIFF
--- a/app/api/logging/log_formatters.py
+++ b/app/api/logging/log_formatters.py
@@ -51,11 +51,17 @@ class HumanReadableFormatter(logging.Formatter):
     def format(self, record: logging.LogRecord) -> str:
         super(HumanReadableFormatter, self).format(record)
 
+        # In certain contexts (like during a request) we want to
+        # attach additional information (like a request ID) to all
+        # log messages for tracking purposes. Add that here.
+        additional_log_context = get_logging_context_attributes()
+        extra = record.__dict__ | additional_log_context
+
         return decodelog.format_line(
             datetime.utcfromtimestamp(record.created),
             record.name,
             record.funcName,
             record.levelname,
             record.message,
-            record.__dict__,
+            extra,
         )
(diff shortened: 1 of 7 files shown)


lorenyu (commented):
@chouinar should I still review this PR, or would we end up changing it anyways based on the decision around openapi framework (switching from connexion)?

chouinar:
> @chouinar should I still review this PR, or would we end up changing it anyways based on the decision around openapi framework (switching from connexion)?

I'd still like to merge this as that switch will only adjust the model definitions. The more endpoints I have to work with, the more I can document how it's different as I switch it over.

lorenyu (approved):

Issue #12, opened 2022-09-20
Look into alternatives to connexion for flask openAPI support


lorenyu:
See [slack thread](https://nava.slack.com/archives/C03G1SWD9H7/p1663703304981579)

Quick google shows a bunch of options, but not sure which is best:
- [flask-rest-api](https://flask-rest-api.readthedocs.io/en/stable/openapi.html)
- [flassger](https://github.com/flasgger/flasgger)
- [flask-openapi3](https://luolingchun.github.io/flask-openapi3/)
- [apiflask](https://apiflask.com/openapi/)

lorenyu:
@chouinar i'm taking a pass at refining the backlog and was trying to remember where we landed on this. do you remember?

chouinar:
This eventually became an ADR I wrote + https://github.com/navapbc/template-application-flask/pull/20 - but we ran into some issues on getting things to be super happy, and it stalled.

lorenyu:
Ah right I remember now. Maybe I'll take a look at that PR to refresh my memory and see if I can pick it back up.

PR #18, opened 2022-09-27
Connexion replacement decision record

chouinar:
## Ticket

https://github.com/navapbc/template-application-flask/issues/12

## Changes
Document describing options to replace our Flask+Connexion approach with Flask+<something>.

## Context for reviewers

Let me know if there are any other libraries that you have come across that might be worth looking into. I did quite a bit of digging, but could certainly be missing others. Also let me know if you want any further details on any of these approaches.

Most of these libraries use [Marshmallow](https://marshmallow.readthedocs.io/en/stable/) for data models which is somewhat similar to Pydantic in being a serialization/validation library meant for APIs & DBs. It has a slightly different format, but most notably you specify Schemas rather than objects with validation. This makes using it slightly annoying, but a library called [Marshmallow Dataclass](https://github.com/lovasoa/marshmallow_dataclass) exists which makes that a non-issue and lets us use dataclasses. There's a brief section in the document under the APIFlask section that shows an example of this, and if we go with that approach, I'd strongly recommend we use Marshmallow Dataclasses.

The super simple tl;dr; of each framework tested:
* APIFlask - Very impressive, great documentation, very customizable, and very intuitive. Seems to have learned from several other frameworks.
* Flask OpenAPI3 - On the surface looks pretty good as it's the only one that works with Pydantic, but documentation is sparse and doesn't actually make a working Swagger endpoint. Managed by one person.
* Flasgger - Annoying to use - basically expected to be used much like Connexion in writing swagger docs directly. The model-based approaches didn't work for me.
* Flask Smorest - Works, but is barebones. Doesn't provide any utility, but does get a Swagger endpoint running from models.
* APISpec - Does not actually vend a Swagger endpoint. Only generates Swagger docs. Would need to be paired with another approach like Flasgger (or Connexion?).
* Flask RESTX - The models are defined as dictionaries, so doesn't actually give us classes to work with. Effectively just the same as defining the swagger spec in code right now. Also appears to be unmaintained.
* APIFairy - Frustrating documentation. No concrete examples of a minimal app made it difficult to get working. APIFlask was a fork of this originally so does a lot of the same things better and clearer.

APIFlask might be the only one I'd actually say is worth the effort of switching off of Connexion as all the others have pretty glaring flaws.

## Testing
The document provides example implementations which each were implemented in order to tinker with the various libraries. Each approach implements a GET/POST endpoint for a simple User object with a few different validations. Each would run as-is alongside installing any relevant library package (eg. `poetry add apiflask`) and running `poetry run python api/route/your_route_file.py`.

Note that not all of them were able to be run - some of them either were very out-of-date and incompatible with recent versions of Flask, or the documentation wasn't detailed enough to solve it quickly for the timebox I put each solution into.

### DIFF
--- a/docs/decisions/0001-connexion-replacement.md
+++ b/docs/decisions/0001-connexion-replacement.md
@@ -1,0 +1,602 @@
+# [short title of solved problem and solution]
+
+* Status: Proposed
+* Deciders: TODO
+* Date: 2022-09-27
+
+Technical Story: https://github.com/navapbc/template-application-flask/issues/12
+
+## Context and Problem Statement
+
+This template library was built out using Flask + Connexion as the API technology. We are investigating if we want to keep using Connexion, or potentially move to another Flask-wrapper API approach. Any approach we consider must make it possible to easily setup an OpenAPI/Swagger endpoint.
+
+We aren't looking to move off of Flask at the moment as we feel that would be a much larger refactor.
+
+## Decision Drivers
+
+* Connexion requires you to specify an OpenAPI spec, but passes the request objects as an the raw JSON. We currently use Pydantic to convert this into a python object for additional validations & for ease-of-use, but this effectively means we need to define any models twice.
+* Defining the OpenAPI specs first instead of defining the data models in code has often been more frustrating. While one goal of doing it this way was the ability to create mock-endpoints while development is ongoing, this can also be easily handled by just making the endpoint return a static response in-code.
+* Connexion's defaults for validation leave a lot to be desired. While we currently override these, it adds a lot of boilerplate to this template library, and isn't immediately obvious.
+
+## Considered Options
+
+* [APIFlask](#apiflask) **Recommended Approach**
+* [Flask OpenAPI3](#flask-openapi3)
+* [Flasgger](#flasgger)
+* [Flask Smorest](#flask-smorest)
+* [APISpec](#apispec)
+* [Flask RESTX](#flask-restx)
+* [APIFairy](#apifairy)
+
+## Decision Outcome
+
+Chosen option: "APIFlask", because it appears to be a library that has taken several lessons from the other libraries. It aims to simplify a lot of the headache with setting up and vending an OpenAPI schema, while being incredibly well documented, and very easy to configure. The documentation does a great job of explaining each component, the rationality of the default behavior, and how to adjust that behavior if needed.
+
+### Positive Consequences <!-- optional -->
+
+* No more writing OpenAPI specs, instead writing models once in-code.
+* TODO - not sure if I actually need to fill this section out, or just say "see: the good things section below"
+
+### Negative Consequences <!-- optional -->
+
+* If we use Marshmallow-dataclass, it's a bit cumbersome to define the validation rules as they all get passed via a metadata map. A utility for generating this so you can instead do `phone_number: str = field(metadata=get_metadata(required=True, regex="...", example="..."))` would go a long way in mitigating this annoyance.
+
+## Pros and Cons of the Options
+
+### APIFlask
+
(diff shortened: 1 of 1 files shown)


docs/decisions/0001-connexion-replacement.md
```diff
@@ -0,0 +1,602 @@
+# Connexion Alternatives Decision Record
```
chouinar, line 1: For your convenience, you can see the formatted document [here](https://github.com/navapbc/template-application-flask/blob/chouinar/connexion-replacement/docs/decisions/0001-connexion-replacement.md).

docs/decisions/0001-connexion-replacement.md
```diff
@@ -0,0 +1,602 @@
...
+Chosen option: "APIFlask", because it appears to be a library that has taken several lessons from the other libraries. It aims to simplify a lot of the headache with setting up and vending an OpenAPI schema, while being incredibly well documented, and very easy to configure. The documentation does a great job of explaining each component, the rationality of the default behavior, and how to adjust that behavior if needed.
+
+### Positive Consequences <!-- optional -->
+
+* No more writing OpenAPI specs, instead writing models once in-code.
+* TODO - not sure if I actually need to fill this section out, or just say "see: the good things section below"
+
+### Negative Consequences <!-- optional -->
+
+* If we use Marshmallow-dataclass, it's a bit cumbersome to define the validation rules as they all get passed via a metadata map. A utility for generating this so you can instead do `phone_number: str = field(metadata=get_metadata(required=True, regex="...", example="..."))` would go a long way in mitigating this annoyance.
```
lorenyu, line 43: can we use apiflask with pydantic dataclasses?

docs/decisions/0001-connexion-replacement.md
```diff
@@ -0,0 +1,602 @@
...
+
+Technical Story: https://github.com/navapbc/template-application-flask/issues/12
+
+## Context and Problem Statement
+
+This template library was built out using Flask + Connexion as the API technology. We are investigating if we want to keep using Connexion, or potentially move to another Flask-wrapper API approach. Any approach we consider must make it possible to easily setup an OpenAPI/Swagger endpoint.
+
+We aren't looking to move off of Flask at the moment as we feel that would be a much larger refactor.
+
+## Decision Drivers
```
lorenyu, line 15: I stumbled across [this article by Swagger itself about design-first vs code-first approaches](https://swagger.io/blog/api-design/design-first-or-code-first-api-development/) and think it might be nice to include in the ADR. Perhaps we can reference it and talk about how in most of our projects we default to speed of development early on in the project, which is why we want something like that for the template project, and that it still allows projects to individually migrate to a design-first approach if they want later on in their project.

docs/decisions/0001-connexion-replacement.md
```diff
@@ -0,0 +1,602 @@
...
+
+
+api.register_blueprint(user_blp)
+if __name__ == "__main__":
+    app.run(port=8080)
+```
+
+</details>
+
+* Good, because the API uses Marshmallow for its data schema, which is a [well maintained library](https://github.com/marshmallow-code/marshmallow) - it's in the same project.
```
lorenyu, line 309: Interesting that it's in the same org as marshmallow. But looks like currently even the maintainer of smorest acknowledges that it has some tech debt (https://github.com/apiflask/apiflask/discussions/14#discussioncomment-571898).

lorenyu (approved):
lgtm from my perspective, would be curious to get others' thoughts who might have more perspective about working with connexion

docs/decisions/0001-connexion-replacement.md
chouinar: Not as far as I can see. It's built ontop of Marshmallow. The only one that used Pydantic at all was Flask OpenAPI3 and it doesn't look to be completed and isn't viable.

PR #20, opened 2022-09-29
Switch out Connexion for APIFlask

chouinar:
## Ticket

{TICKET_LINK} - TODO

## Changes

This is a draft as a few areas of implementation might warrant a bit of discussion (will comment on the PR for areas I'd like to talk through).

This replaces Connexion with APIFlask, a separate framework for wrapping Flask that generates API models

## Context for reviewers
This approach specifically uses Marshmallow Dataclasses instead of just Marshmallow (as detailed at https://apiflask.com/schema/#use-dataclass-as-data-schema ). This approach lets us always work with just dataclass objects which auto-generate the marshmallow schema behind the scenes.

## Testing
TODO - add a few screenshots.

You can run locally via `make clean-volumes init run-logs` and go to `localhost:8080/docs` to see the endpoints.


app/api/route/error_handlers.py
```diff
@@ -1,166 +1,27 @@
...
-        err_message = e["msg"]
-
-        errors.append(
-            ValidationErrorDetail(
-                type=err_type,
-                message=f'Error in field: "{err_field}". {err_message.capitalize()}.',
-                rule=None,
-                field=".".join(str(loc) for loc in e["loc"]),
-            )
+def add_error_handlers_to_app(app: APIFlask) -> None:
```
chouinar, line 11: APIFlask has much saner error formatting than Connexion (ie. multiple validation errors gives you all the errors). I only kept this to make clear you can adjust the formatting if desired.

app/api/route/models/param_schema.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+@dataclass
+class ParamFieldConfig:
+    example: Optional[str | bool | int] = None
+    description: Optional[str] = None
+    allowed_values: Optional[Iterable | Type[Enum]] = None
+    regex: Optional[str] = None
+
+    def build(self, **overrides: Any) -> Any:
```
chouinar, line 20: This whole ParamFieldConfig class is attempting to make building the configuration for examples/descriptions/regexes/etc. a lot simpler, but let me know if you have any suggestions to organize this differently.

For example, lets say you want to validate a phone number has the right format, and provide a default. If you do that directly in your model, you'd need to do:

```py
...
    phone_number = field(metadata={
        "metadata":{"example":"123-456-7890", "description":"it's a phone number!"},
        "validate": [Regexp("... your regex")]
    }
```
Which is a little oddly formatted, as you've got two metadata objects which contain different info. To make it simpler, this util makes it so you can just pass in simple params which will auto-convert to that.

Additionally some fields probably want the exact same config approach (`created_at` and `updated_at` are both just datetime), so we can share that config, which is what the ParamSchema class below is doing. The build function itself takes in overrides for some fields in case you want to use say `first_name` and make the description more specific.

Note that we cannot do:
```py
class ParamSchema:
     datetime = field(...)

@marshmallow_dataclass
class MyModel:
    created_at = ParamSchema.datetime
    updated_at = ParamSchema.datetime
```
as if you have multiple fields reference the same call to the field function, it just isn't happy and only displays one in Swagger (guessing some sort of binding underneath the hood?).

app/api/route/request.py
```diff
@@ -1,27 +1,39 @@
...
-        orm_mode = True
+logger = api.logging.get_logger(__name__)
 
-    def get_set_params(self) -> Iterable[tuple[str, Any]]:
+PARAM_UNSET_IN_REQUEST: str = "PARAM_UNSET_IN_REQUEST"
+
+
+@marshmallow_dataclass
+class BaseRequestModel(BaseApiModel):
+    def get_set_params(self, api_context: ApiContext) -> Iterable[tuple[str, Any]]:
```
chouinar, line 15: This is easily the most frustrating thing I ran into. For an update function, we ONLY want to set parameters that the user passed in. So if you do `PATCH` and only set `first_name` we should only work with that. Pydantic was useful in that it kept track of the values passed into its init, but dataclasses don't do that. For dataclasses, we have two options:

1. In the request, we can set `UserPatchParams.Schema(partial=True)` which will call the dataclass with only the params passed in. But for this we need default values for every init variable so the dataclasses init function gets generated as `def __init__(self, param1 = <default>, param2 = <default>, ...)`. I tried using this approach and forcing all the defaults to be `PARAM_UNSET_IN_REQUEST` but this breaks the swagger endpoint for any non-strings during serialization. So this approach doesn't work
2. We allow the dataclass to be initialized with values defaulting to `None` and then refer back to the raw API request object from Flask to see what was actually passed in. This works, but any nested params aren't going to work well here. Although that might just make you split your patch methods up as is recommended.

Open to any alternative ideas.

lorenyu:
@chouinar double checking, did you want an initial review on this while it's in draft, or should i wait until it's in "ready for review" state?

chouinar:
> double checking, did you want an initial review on this while it's in draft, or should i wait until it's in "ready for review" state?

This is ready for review - the draft bit is to make clear there are a few areas I expect discussion + I need to update the docs / ticket description a bit more. Functionally it should be complete.

app/Makefile
```diff
@@ -127,7 +126,7 @@ format:
 	$(PY_RUN_CMD) isort --atomic api tests
 	$(PY_RUN_CMD) black api tests
 
-lint: lint-spectral lint-py
```
lorenyu, line 130: for my own understanding, is the idea that we don't need to lint the openapi spec since it's autogenerated?

app/api/app.py
```diff
@@ -89,9 +66,65 @@ def current_user(is_user_expected: bool = True) -> Optional[User]:
...
         logger.error("No current user found for request")
-        raise Unauthorized
+        # raise Unauthorized
     return current
 
 
+def configure_app(app: APIFlask) -> None:
+    # Modify the response schema to instead use the format of our Response class
+    # which adds additional details to the object.
+    app.config["BASE_RESPONSE_SCHEMA"] = marshmallow_dataclass.class_schema(Response)
```
lorenyu, line 76: can we put the schema next to the dataclass e.g.

```
from api.route.response import Response, ResponseSchema
```

then this line would become

```
app.config["BASE_RESPONSE_SCHEMA"] = ResponseSchema
```

that way we keep the schema's close to the dataclass and also clean up the dependencies on marshmallow_class, ideally only the data model files have a direct dependency on the validation library.

or if i'm understanding the rest of this PR correctly, would it be more accurate to do

```
app.config["BASE_RESPONSE_SCHEMA"] = Response.Schema
```

app/api/app.py
```diff
@@ -89,9 +66,65 @@ def current_user(is_user_expected: bool = True) -> Optional[User]:
...
+
+    # Set a few values for the Swagger endpoint
+    app.config["OPENAPI_VERSION"] = "3.0.3"
```
lorenyu, line 79: curious what this does. maybe link to the apiflask doc that explains what this config does? does it change how the openapi spec is autogenerated?

app/api/app.py
```diff
@@ -89,9 +66,65 @@ def current_user(is_user_expected: bool = True) -> Optional[User]:
...
+    # Set various general OpenAPI config values
+    app.info = {
+        "title": "Template Application Flask",
+        "description": "Template API for a Flask Application",
+        "contact": {
+            "name": "Nava PBC Engineering",
+            "url": "https://www.navapbc.com",
+            "email": "engineering@navapbc.com",
+        },
+    }
```
lorenyu, line 90: what's the difference between `app.config["INFO"] = `and `app.info = `? if there's no difference could we use `app.config["INFO"]` to be consistent with the previous configuration settings, or conversely change the other ones to use the attribute method like `app.open_api_version`?

if one of them gives us better type safety then maybe we can use that one (e.g. if the attributes are defined at compile-time then that might could be checked by mypy)

lorenyu: what do you think about leaving in placeholder values like
```suggestion
    app.info = {
        "title": "<APPLICATION_NAME>",
        "description": "<APPLICATION_DESCRIPTION>",
    }
```

app/api/app.py
```diff
@@ -2,15 +2,17 @@
 from contextlib import contextmanager
 from typing import Generator, Optional
 
-import connexion
+import marshmallow_dataclass
+from apiflask import APIFlask
 from flask import g
```
lorenyu, line 7: does `apiflask` not re-export `g`?

app/api/auth/api_key_auth.py
```diff
@@ -4,12 +4,22 @@
...
-from werkzeug.exceptions import Unauthorized
+from apiflask import HTTPTokenAuth, abort
 
 import api.logging
 
 logger = api.logging.get_logger(__name__)
 
+# Initialize the authorization context
+# this needs to be attached to your
+# routes as `your_blueprint..auth_required(api_key_auth)`
```
lorenyu, line 15:
```suggestion
# routes as `your_blueprint.auth_required(api_key_auth)`
```

app/api/app.py
```diff
@@ -89,9 +66,65 @@ def current_user(is_user_expected: bool = True) -> Optional[User]:
...
+            "name": "Nava PBC Engineering",
+            "url": "https://www.navapbc.com",
+            "email": "engineering@navapbc.com",
+        },
+    }
+
+    # Set the security schema and define the header param
+    # where we expect the API token to reside.
+    # See: https://apiflask.com/authentication/#use-external-authentication-library
+    app.security_schemes = {"ApiKeyAuth": {"type": "apiKey", "in": "header", "name": "X-Auth"}}
```
lorenyu, line 95: seems like you define this in `get_app_security_scheme` in `api_key_auth`, should we be using that function?

app/api/route/error_handlers.py
```diff
@@ -1,166 +1,27 @@
...
+        # and exists as a template for if you wish to modify
+        # the format of error responses.
+        # Note that if you change this format, you'll
+        # also need to specify the error schema so
+        # OpenAPI will provide the correct format:
+        # https://apiflask.com/error-handling/#update-the-openapi-schema-of-error-responses
+        return (
+            {"status_code": error.status_code, "message": error.message, "detail": error.detail},
+            error.status_code,
+            error.headers,
```
lorenyu, line 26: this is a good intention. that said if it's not needed i lean towards cutting for now. i think there could be cognitive overload and it would be nice if the template was as close to ready to go to prod as possible, so i lean towards minimizing no-op boiler plate code.

app/api/route/handler/user_handler.py
```diff
@@ -66,14 +32,21 @@ def create_user(api_context: ApiContext) -> UserResponse:
...
+        # Unfortunately there isn't a great way of determining
+        # which values were explicitly set in the request other
+        # than to look at the raw request JSON. Because the objects
+        # get converted to dataclasses, all the values are "set" and
+        # dataclasses don't keep track of the parameters
+        if key not in api_context.request_body:
+            continue
 
         if key == "roles":
             handle_role_patch(user, value, api_context)
```
lorenyu, line 52: not necessarily in scope for this PR, but do we need to handle patching roles? that seems extra for the template, i imagine most of projects won't provide the ability for the api client to modify a user's role. i imagine the role would be set when the user is created. if we don't need it i think we could simplify the template by cutting it. thoughts?

app/api/route/request.py
lorenyu: > In the request, we can set UserPatchParams.Schema(partial=True) which will call the dataclass with only the params passed in. But for this we need default values for every init variable so the dataclasses init function gets generated as def __init__(self, param1 = <default>, param2 = <default>, ...).

could you elaborate on that more? [the docs on partial loading in marshmallow](https://marshmallow.readthedocs.io/en/stable/quickstart.html#partial-loading) seem to indicate that the resulting object won't even have the missing fields, so we don't need default values for them:

relevant snippet
```
class UserSchema(Schema):
    name = fields.String(required=True)
    age = fields.Integer(required=True)

result = UserSchema().load({"age": 42}, partial=True)
# OR UserSchema(partial=True).load({'age': 42})
print(result)  # => {'age': 42}
```

app/api/route/models/base_api_model.py
```diff
@@ -0,0 +1,12 @@
...
+from marshmallow import Schema  # noqa: F401
+from marshmallow_dataclass import dataclass as marshmallow_dataclass
+
+
+@marshmallow_dataclass
+class BaseApiModel:
+    # Adding this to the base model helps
+    # mypy recognize the .Schema that is dynamically
+    # added by marshmallow_dataclass
+    Schema: ClassVar[Type[Schema]] = Schema  # noqa: F811
```
lorenyu, line 12: i took a look at [how marshmallow dataclasses implements adding the `Schema` attribute to the class object](https://github.com/lovasoa/marshmallow_dataclass/blob/master/marshmallow_dataclass/__init__.py#L212-L219), it looks like in their type hint they are not actually recognizing that the return value from the decorator is a different type with an extra attribute:
```
def decorator(clazz: Type[_U]) -> Type[_U]:
```
IMO i think this is bad design by the marshmallow dataclass library. instead of using `Model.Schema`. let's instead have the schema be separate, like this:

```
# user
import marshmallow_dataclass

class User:
  ...

UserSchema = marshmallow_dataclass.class_schema(User)
```

then in places that need the schema, they'll import `UserSchema`

lorenyu (commented):
sorry for the delay! did a first review pass!

app/api/route/request.py
chouinar: There's two ways to use Marshmallow, as a schema, or as dataclasses which "wrap" the schema.

If you use just the straight Schema approach, the request body objects you work with are dictionaries/JSON. They've been validated, and types will be converted (eg. a date string to a date object), but you'll have more typing headaches. 

With marshmallow-dataclasses, it does the same Schema validation as before, but then calls your dataclass to actually give you something better to work with. It's effectively the same as doing:
```py
dataclass_obj = YourDataclass(**request_json)
```
at the start of each of your request methods, but handling nested objects a bit better.

You can tell it to use `partial=True` with marshmallow dataclasses, BUT if you do that without any defaults, you're calling the `__init__` function of your dataclass with only a few required parameters, which will error. If you give those parameters defaults (eg. `__init__(self, x = None, y = None)`) then you can't tell what was actually set, and what wasn't. I tried various ways to set defaults that made it obvious what was set, but then you need to handle every separate type (for strings putting `THIS_WAS_NOT_SET` works, but what do you do for dates - the value has to be valid).

app/api/route/models/base_api_model.py
chouinar: Yeah, that seems fair. Might help to distinguish that there are in fact TWO classes for every model here, the auto-generated Schema, and the dataclass you work with.

app/Makefile
chouinar: Right, the main thing this validated was human-error in generating the file / helping figure out OpenAPIs obtuse errors complaining about the format of the file. We could in theory have it validate the spec generated, but not sure if that has a ton of value.

app/api/route/request.py
lorenyu: I see, that makes sense. The tricky thing for `patch` requests is that there are many possible classes/object types, involving different combinations of attributes. Based on this discussion, what about having the request params / responses use the straight Schema approach, and only convert to the object type once we get past the request handling layer. For example, unvalidated request body --> validate with request params schema which produces a request params dict --> send to `api.services.user_service.patch_user(patch_data: Dict)` which has methods that take validated inputs.

app/api/app.py
chouinar: Good catch, meant to have that defined in the auth code, and forgot to call the method itself.

app/api/route/handler/user_handler.py
chouinar: The way patch works for a user is already a bit iffy. Roles are handled, but it has a lot of complexity (see `handle_role_patch`). Based on what I've read online, patch endpoints probably make sense to split up for anything dealing with one-to-many relationships. Roles should probably be handled by endpoints like `PATCH /user/:id/add-role` and `PATCH /user/:id/remove-role/:role_id` rather than this way. Those would work significantly more intuitively and avoid all of the complexity. It just leads to having a lot more endpoints potentially.

lorenyu: agree that patching roles this way is iffy. shall we remove the patching of roles from the template for now then?

app/api/route/request.py
chouinar: You might be right. I'd like to keep dataclasses, but let's see what the effort would be to have the endpoints based purely on the schema/json, even if it means we need to do `MyDataclass(**request_json)` at the start of every request then so be it if it simplifies the patch approach elsewhere.

lorenyu: sounds good!

chouinar: I can't seem to figure out a great way to do that unfortunately. If I define just dataclasses and then convert them, it doesn't handle nested objects/inheritance very well without several workarounds there.

I might be able to inject the info into the dataclass while it loads, but even that is proving to be a bit hacky.

The "clean" way to do it is to define a Schema, and separately define a dataclass, but then we're duplicating every model, and I can't figure out a way to dedupe these.

app/api/route/models/base_api_model.py
chouinar: Similar to the issue described elsewhere, I'm not sure I can actually do this and maintain inheritance. If I define a `@dataclass` for the model then it runs into several issues with fields being defined out of order regarding the defaults. There might be a way to work around this, but I'm not sure I have the bandwidth to try and restructure that at this time.

app/api/route/request.py
lorenyu: not handling inheritance should be ok, i think we can make do without inheritance for data models.

not handling nested objects seems like a bigger issue. could you elaborate on what you mean by that? if you have classes like

```python
@dataclass
class UserRole:
  ...

@dataclass
class User:
  name: str
  role: UserRole
```

then do `UserSchema = class_schema(User)`, what happens in that situation? apologies if i'm not understanding the issue properly.

chouinar:
```py
@dataclass
class UserRole:
  description: str

@dataclass
class User:
  name: str
  roles: list[UserRole]
```
When you use this as the basis for the Schema, it doesn't add the UserRole to the schema for some reason and calling with `{"name":"bob", "roles": [{..anything.}]}` errors as the field is unknown to it.

lorenyu: oh. that's odd, the example on [their docs](https://pypi.org/project/marshmallow-dataclass/) seems to handle that use case:

```python
from dataclasses import dataclass, field
from typing import List, Optional

import marshmallow_dataclass
import marshmallow.validate


@dataclass
class Building:
    # field metadata is used to instantiate the marshmallow field
    height: float = field(metadata={"validate": marshmallow.validate.Range(min=0)})
    name: str = field(default="anonymous")


@dataclass
class City:
    name: Optional[str]
    buildings: List[Building] = field(default_factory=list)


city_schema = marshmallow_dataclass.class_schema(City)()

city = city_schema.load(
    {"name": "Paris", "buildings": [{"name": "Eiffel Tower", "height": 324}]}
)
# => City(name='Paris', buildings=[Building(height=324.0, name='Eiffel Tower')])

city_dict = city_schema.dump(city)
# => {'name': 'Paris', 'buildings': [{'name': 'Eiffel Tower', 'height': 324.0}]}
```

i wonder if it's because we need the `field(default_factory=list)` ?

chouinar: Maybe? We need to specify `field(...)` otherwise when the dataclass generates, it complains that you've specified default and non-default `__init__` params out of order. Strangely, the `@marshmallow_dataclass` decorator handles this better and knows if we don't specify something, to properly recurse and use the whole nested class.

chouinar:
Closing as @lorenyu handled this in https://github.com/navapbc/template-application-flask/pull/51

PR #21, opened 2022-09-29
Add API checks for PRs and merge to main

### LINKED ISSUE #19: Add CI
# Context

template-application-flask still doesn't have CI. We want to add a `ci-app.yml` that runs the checks for the flask application. This CI should be able to be used on any project that uses the template.

See this [github thread](https://github.com/navapbc/template-infra/pull/106#discussion_r982855881)

chouinar:
## Ticket

Resolves #19 

## Changes

Added CI for validating linting and tests on PR + merge to main.

## Context for reviewers

Let me know if the make commands should be merged or renamed based on what you were thinking of regarding the abstract Makefile approach.

## Testing
You can see the runs from this particular PR: https://github.com/navapbc/template-application-flask/actions/workflows/api-checks.yml

### DIFF
--- a/.github/workflows/api-checks.yml
+++ b/.github/workflows/api-checks.yml
@@ -1,0 +1,55 @@
+name: Template API Checks
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - app/**
+      - .github/workflows/api-checks.yml
+  pull_request:
+    paths:
+      - app/**
+      - .github/workflows/api-checks.yml
+
+defaults:
+  run:
+    working-directory: ./app
+
+env:
+  # Some commands in the Makefile check if this is set
+  # and add additional parameters to the commands called
+  CI: true
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: build container
+        run: make build
+
+      - name: Run format check
+        run: make format-check
+
+      - name: Run linting
+        run: make lint
+  security:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Security Check
+        run: make lint-security
+  test:
+    needs: [lint, security]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: test for multi-headed migration situation
+        run: make test args="-x tests/api/db/test_migrations.py::test_only_single_head_revision_in_migrations" 
+
+      - name: Start tests
+        run: make test-coverage
--- a/.github/workflows/ci-app.yml
+++ b/.github/workflows/ci-app.yml
@@ -16,18 +16,15 @@ defaults:
   run:
     working-directory: ./app
 
-env:
-  # Some commands in the Makefile check if this is set
-  # and add additional parameters to the commands called
-  CI: true
+
 
 jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
-      - name: build container
+      - name: Build container
         run: make build
 
       - name: Run format check
@@ -35,21 +32,19 @@ jobs:
 
       - name: Run linting
         run: make lint
-  security:
+
+  lint-security:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
-      - name: Security Check
+      - name: Run security linting
         run: make lint-security
+
   test:
-    needs: [lint, security]
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-
-      - name: test for multi-headed migration situation
-        run: make test args="-x tests/api/db/test_migrations.py::test_only_single_head_revision_in_migrations" 
+      - uses: actions/checkout@v3
 
       - name: Start tests
         run: make test-coverage
--- a/.github/workflows/ci-app.yml
+++ b/.github/workflows/ci-app.yml
@@ -16,16 +16,11 @@ defaults:
   run:
     working-directory: ./app
 
-
-
 jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-
-      - name: Build container
-        run: make build
 
       - name: Run format check
         run: make format-check
--- a/.github/workflows/ci-app.yml
+++ b/.github/workflows/ci-app.yml
@@ -17,7 +17,11 @@ defaults:
     working-directory: ./app
 
 jobs:
-  lint:
+  lint-test:
+    # We could run these steps in parallel
+    # if we built the docker image and shared
+    # it across jobs as described in:
+    # https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#passing-data-between-jobs-in-a-workflow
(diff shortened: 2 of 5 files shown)


.github/workflows/api-checks.yml
```diff
@@ -0,0 +1,55 @@
+name: Template API Checks
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - app/**
+      - .github/workflows/api-checks.yml
```
lorenyu: Can we rename the workflow to `ci-app.yml`

.github/workflows/api-checks.yml
```diff
@@ -0,0 +1,55 @@
...
+
+defaults:
+  run:
+    working-directory: ./app
+
+env:
+  # Some commands in the Makefile check if this is set
+  # and add additional parameters to the commands called
+  CI: true
```
lorenyu: i don't believe this is necessary, since [according to github docs, github actions always sets CI to true](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables)

.github/workflows/api-checks.yml
```diff
@@ -0,0 +1,55 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
lorenyu: not sure the difference but there's a v3 now so might as well use that here and in the rest of the file too

```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/api-checks.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Security Check
+        run: make lint-security
+  test:
+    needs: [lint, security]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: test for multi-headed migration situation
+        run: make test args="-x tests/api/db/test_migrations.py::test_only_single_head_revision_in_migrations"
```
lorenyu: curious, does this test not run as part of `test-coverage`? why is it called out separately?

.github/workflows/api-checks.yml
```diff
@@ -0,0 +1,55 @@
...
+    needs: [lint, security]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: test for multi-headed migration situation
+        run: make test args="-x tests/api/db/test_migrations.py::test_only_single_head_revision_in_migrations" 
+
+      - name: Start tests
+        run: make test-coverage
```
lorenyu: nit: missing newline at end of file

.github/workflows/api-checks.yml
```diff
@@ -0,0 +1,55 @@
...
+  # and add additional parameters to the commands called
+  CI: true
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: build container
```
lorenyu: nit: use sentence case to be consistent with other step names
```suggestion
      - name: Build container
```

.github/workflows/api-checks.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: make lint
+  security:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Security Check
+        run: make lint-security
+  test:
+    needs: [lint, security]
```
lorenyu: Is this necessary? Seems like we could potentially `test` at the same time as `lint` and `security`

.github/workflows/api-checks.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: make format-check
+
+      - name: Run linting
+        run: make lint
+  security:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Security Check
```
lorenyu: maybe rename to `Run security linting` to be more specific. I think there are other tools that do more in depth security checks like static code analysis

.github/workflows/api-checks.yml
```diff
@@ -0,0 +1,55 @@
...
+  CI: true
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: build container
+        run: make build
```
lorenyu: Since we also split out `security` and `test` could we move `build` to a separate job? I think of checking that the build build works as different from linting.

lorenyu (approved):
Looks great thanks!

Question: are we running mypy type checking anywhere? If not I'll add it to the backlog

.github/workflows/api-checks.yml
chouinar: Looks like just an upgrade to the node version: https://github.com/actions/checkout

.github/workflows/api-checks.yml
chouinar: Not exactly - technically this could be removed. Every one of these actually runs the same docker build as they each recreate the docker container in order to actually run. If you look at the logs, you can see that they build and fetch packages first, which is why tests takes ~2minutes despite the actual tests running in about 10 seconds. there is a more complex way of building once and reusing across tasks not implemented here

.github/workflows/api-checks.yml
chouinar: Let me test that

.github/workflows/api-checks.yml
chouinar: It does. I think the idea in the code it's based off of had to do with when the migrations ran, although from my testing that didn't matter (maybe just a legacy artifact of a much older approach?). I'll remove it.

.github/workflows/api-checks.yml
chouinar: Should we add a tool for linting/formatting of yaml files to the backlog?

.github/workflows/api-checks.yml
chouinar: I agree, not sure if the dependency helps

chouinar:
> Question: are we running mypy type checking anywhere? If not I'll add it to the backlog

It runs as part of the `make lint` command along with flake.

.github/workflows/api-checks.yml
lorenyu: oh i see. in that case to keep things simple why don't we just run all the steps in one job so that they can all reuse the same build? and we can create a ticket in the future to [reuse build artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts).

.github/workflows/api-checks.yml
lorenyu: 👍 , created #25

.github/workflows/api-checks.yml
lorenyu: actually nvm, if running them in parallel is faster let's keep it that way, your call!

lorenyu:
> [type checking] runs as part of the make lint command along with flake.

ah ok thanks for explaining that. more of a nit and not necessary for this PR, but i think of type-checking as a more advanced form of static analysis / static checks than linting, so i conceptually think of it as a separate thing. that said, it's just mincing semantics so no biggie either way!

.github/workflows/api-checks.yml
chouinar: It's not much different since 80% of each of them is just the build - this way is probably more logical until we switch it over.

PR #24, opened 2022-09-30
Add release build target

lorenyu:
## Ticket

n/a

## Changes
- Add `release` build target to Dockerfile

## Context for reviewers
In this [PR template-infra#116](https://github.com/navapbc/template-infra/pull/116) we established an interface such that every application's Dockerfile should have a named build stage called `release`, so that `make release-build` from the template-infra `Makefile` can integrate properly with the application's Dockerfile.

## Testing
Didn't test, since I just added a name to the build stage but didn't change anything else


chouinar (approved):
LGTM!

PR #31, opened 2022-12-12
Fix build of pyscopg2 wheel during poetry install

### LINKED ISSUE #28: Setup instructions fail if postgresql not installed locally
# Context

Image

Looks like installing postgresql fixes it

https://stackoverflow.com/questions/11618898/pg-config-executable-not-found

lorenyu:
## Ticket

Resolves #28 

## Changes
* Install postgresql and libpq in Docker image
* Add getting started instructions

## Context for reviewers
Might need some more PRs to clean things up more (e.g. the getting started instructions sets things up locally but then we end up running in Docker which doesn't even use the local installation instructions). But I thought for now I'd fix the bug in the Dockerfile


app/Dockerfile
```diff
@@ -5,13 +5,16 @@
...
+
 # Install poetry, the package manager.
 # https://python-poetry.org
 RUN pip install poetry
 
+RUN apt-get update \
+    && apt-get install --no-install-recommends --yes \
+    build-essential \
+    libpq-dev \
+    postgresql
```
chouinar, line 17: I'm guessing you had to add this because of something in the python 3.11 image. We ran into a similar issue with the oracle libraries in 3.11 where the slim image suddenly was missing libraries.

chouinar (approved):
LGTM!

app/Dockerfile
lorenyu: Oh interesting. Btw I noticed on another project we use slim-bullseye. Do you know the rationale behind that

PR #35, opened 2022-12-13
Do not specify platform in docker-compose

### LINKED ISSUE #34: Speed up docker builds locally for M1 macs

lorenyu:
## Ticket

Resolves #34 

## Changes
see title

## Context for reviewers
Building the docker image locally is painfully slow on M1 macs. This PR fixes that

## Testing
Before docker-compose build takes > 6 min
image

Now it take < 1 min
image

Healthcheck still runs properly after `make start`
image


chouinar (commented):
I'm on an m1 mac, and I'm not seeing any difference in performance with/without the platform specified. My runtimes are comparable to your slower image. Tested this on both this repo, and the NJ UI API.

Is there something else potentially going on that made it faster for you the second time? I see you specified no-cache, but it was faster with installing libraries that might get cached somewhere anyways?

chouinar:
> I'm on an m1 mac, and I'm not seeing any difference in performance with/without the platform specified. My runtimes are comparable to your slower image. Tested this on both this repo, and the NJ UI API.
> 
> Is there something else potentially going on that made it faster for you the second time? I see you specified no-cache, but it was faster with installing libraries that might get cached somewhere anyways?

Testing my own theory, and rerunning the build, it's still just as slow. Perhaps there is something else at play here in how our laptops are setup? I'm running 12.5 (Monterey), and my laptop is from late-2021.

It took 7:41 & 7:47 respectively.

Perhaps it might be good to pair to find the difference?

lorenyu:
Yeah let's pair, my mac seems to be similar 
image

I'll find some time. In the meantime what are your docker desktop settings? These are mine
image

chouinar (approved):
LGTM!

I synced up with Loren. This does speed things up, the issue was that I was running colima for something in the NJ project which effectively makes Docker think you're running on a different architecture because that's necessary for the DB we use. Turning that off, I saw the build improvements just the same.

PR #41, opened 2022-12-14
Set port from environment variable

### LINKED ISSUE #40: Use PORT env var to set the port that the server listens to
# Context

This is important to ensure that it works with template-infra which sets that env var in ECS

lorenyu:
## Ticket

Resolves #40 

## Changes
* Ignore VScode files
* Set server port from environment variable

## Context for reviewers
Template infra will set PORT env var so we want to use that rather than hardcoding it

## Testing
CI


app/api/__main__.py
```diff
@@ -21,14 +23,15 @@ def main() -> None:
     logger.info("Running API Application")
 
     app = api.app.create_app()
+    port = os.getenv("PORT", 8080)
```
chouinar: Could we put this env var in the app config object?

(I meant to do this when moving it over, but looks like I forgot)

lorenyu: Done!

zelgadis (approved):
LGTM!

Issue #16, opened 2022-09-22
Consider removing dependency on ORM


lorenyu:
# Context

In the main comment thread of https://github.com/navapbc/wic-mt-demo-project-mock-api/pull/20 we discussed not having ORM be in the default template.

lorenyu:
upon further thought i think we can ice this for now

PR #42, opened 2022-12-14
Add db adapter using psycopg directly

lorenyu:
## Ticket

Makes progress on #16 

## Changes
* Add db adapter using psycopg
* Remove db session context manager
* Remove db from Flask request context

## Context for reviewers
I want to slowly move away from SQLAlchemy which I think is more complicated than needed. This PR adds an adapter library for the db using psycopg 3, and uses it in the healthcheck library. It also removes the db session context manager, I'm not convinced that's needed. If we design our code properly, only a couple of places should need the db connection – at the beginning and end of a handler – and we could pass the connection in directly in those places, and we shouldn't make it too easy to access the db connection throughout the codebase.

I didn't completely remove sqlalchemy yet, was gonna consider doing that in a future PR, didn't want to conflict with @zelgadis 's work

## Testing
Tested the healthcheck locally
image

Turned off the db and saw that healthcheck returns 503
image

Turned the db on again and saw that it reconnected after a few hits
image

### DIFF
--- a/app/api/adapters/db/__init__.py
+++ b/app/api/adapters/db/__init__.py
@@ -3,6 +3,17 @@ Database adapter for Postgres database.
 This module is an abstraction around the psycopg3 python postgres library. It
 has methods for connecting to the database, and abstracts away the loading of
 config and environment variables.
+
+Usage:
+    # During application initialization
+    from api.adapters import db
+    db.init()
+
+    # In a route handler
+    from api.adapters import db
+    with db.get_connection() as conn:
+        with conn.cursor() as cur:
+            cur.execute("SELECT 1")
 """
 
 import atexit
@@ -68,6 +79,11 @@ def get_connection() -> psycopg.Connection:
     return pool.connection()
 
 
+# Global connection pool
+# The simplest way to use the psycopg pool is to create a single instance of it,
+# and to use this object in the rest of the program. The pool is threadsafe and
+# can be used by multiple threads at the same time.
+# https://www.psycopg.org/psycopg3/docs/advanced/pool.html
 pool: psycopg_pool.ConnectionPool | None = None
 
 __all__ = ["init", "get_connection", "Connection", "Cursor", "kwargs_row"]
--- a/app/api/adapters/db/__init__.py
+++ b/app/api/adapters/db/__init__.py
@@ -20,8 +20,8 @@ import atexit
 import os
 
 import psycopg
-from psycopg import rows
 import psycopg_pool
+from psycopg import rows
 
 from api import logging
 
--- a/app/api/adapters/db/__init__.py
+++ b/app/api/adapters/db/__init__.py
@@ -1,0 +1,73 @@
+"""
+Database adapter for Postgres database.
+This module is an abstraction around the psycopg3 python postgres library. It
+has methods for connecting to the database, and abstracts away the loading of
+config and environment variables.
+"""
+
+import atexit
+import os
+
+import psycopg
+from psycopg import rows
+import psycopg_pool
+
+from api import logging
+
+# Re-export these types to restrict the API of this module
+# to only the types we want to expose explicitly.
+Connection = psycopg.Connection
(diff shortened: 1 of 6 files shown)


lorenyu:
@chouinar fyi tagging jesse and ellery on PRs now since i think they have some spare time in december

app/api/adapters/db/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+    atexit.register(pool.close)
+
+
+def get_conninfo() -> str:
+    """Gets a db connection info based on environment variables"""
+    dbhost = os.getenv("DB_HOST")
+    dbname = os.getenv("POSTGRES_DB")
+    username = os.getenv("POSTGRES_USER")
+    password = os.getenv("POSTGRES_PASSWORD")
+    ENVIRONMENT = os.getenv("ENVIRONMENT", "production")
```
chouinar, line 57: Would we want a default to be production?

app/api/adapters/db/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+
+def get_connection() -> psycopg.Connection:
+    """Gets a db connection
+    Uses database credentials in environment variables to obtain a database
+    connection.
+    Usage:
+      with get_connection() as conn:
+        with conn.cursor() as cur:
+          cur.execute("SELECT 1")
+          result = cur.fetchall()
```
chouinar, line 77: The security linting tool that runs (Bandit) isn't going to like this, it doesn't let you write raw-sql with any params passed into the text. Will either need to replace it, or ignore all of those warnings which are in place to prevent any sort of SQL injection. Might want some sort of util for making SQL.

I'd also push back on any approach that doesn't come with at least some guidance on how to deserialize a response into something we can work with, the primary benefit of SQLAlchemy in my mind. It's one thing to do `select id from my_table` another to do `select a, b, c, d, e, f, g ... from my_table`.

app/api/adapters/db/__init__.py
chouinar, line 57: Rather than os.getenv, could make a model derived from `PydanticBaseEnvConfig` which loads it for you.

chouinar (commented):
I don't think you'll be able to get this running without all the tests failing unless you fully switch off SQLAlchemy as their entire setup is very tightly coupled (migrations to set up a schema, using the existing DB session). I am curious about how you'll handle migrations as my quick look online when you brought this idea up before gave little to go on.

chouinar:
@lorenyu - Might I suggest an alternative approach - probably worth discussing synchronously, but what do you think of a hybrid approach that uses both SQLAlchemy & low-level SQL as you're adding? SQLAlchemy could be used for defining the DB schema, handling migrations, and very obvious direct CRUD operations, but for complex queries, especially backend processes, we can go a bit more low-level.

No idea if this would be more or less confusing. I'm worried that switching off of an ORM entirely might.. not go great. I wanted to do it in the NJ project but the size of the schema was so massive that working with raw SQL was going to be way too difficult to manage, and too error prone. With some sort of implementation involving examples of binding objects to a response I'd maybe feel better, but I can't imagine wanting to use raw SQL for my current project.

Edit: Looks like binding to pydantic is pretty simple actually - https://www.psycopg.org/psycopg3/docs/advanced/typing.html#example-returning-records-as-pydantic-models - that type of approach would be really nice actually.

lorenyu:
@chouinar yeah great points. I'll need to think about it more
psycopg's class_row or kwargs_row can do the conversion to object models on reads, so I'm less worried about reads, but I think inserts and updates can still be annoying to have to code up for simple ones.

re: migrations, I've been wondering whether migrations should live within the `/app` or at the project root next to infra as it's own `/db`. I was considering using something like https://github.com/amacneil/dbmate that's framework agnostic and just uses sql.

In any case I'm gonna rethink some things and go back to the drawing board. I'm slowly changing my mind about using SqlAlchemy as long as we organize the actual DB interaction logic inside persistence classes

PR #46, opened 2022-12-16
Make naming and typing refinements to db module

lorenyu:
## Changes
* Simplify migration filename template
* Use id instead of user_id for primary key column name
* Create IdMixin to add primary key for models
* Use SQLAlchemy's Enum type, which translates to VARCHAR in PostgreSQL
* Rename RoleEnum to RoleType
* Rename role_description to type
* Rename UserRole to Role
* Use utcnow directly
* use uuid4 directly
* Recreate migration file
* Remove "Third party" option from enum
* Ignore .envrc file used by direnv

## Context for reviewers
This PR replaces the work in #42, rather than moving off SQLAlchemy, just make some small tweaks to simplify the example a bit. Mostly renaming things, including changing primary key to just `id` instead of `[table]_id`, which makes implementing an `IdMixin` simpler and I don't think there's any confusion about what type of id it is since `user.id` is just as clear if not clearer than `user.user_id`.

Also changed to use SQLAlchemy's Enum type with native=False which uses VARCHAR under the hood, I think this gives us better type safety since SQLAlchemy will autoconvert to the enum type RoleType rather than keeping it as a string.

I recreated the migration file as this is a template I didn't want to burden new projects with too many migrations, so people working on the template itself may just want to blow away their DB.

## Testing
CI


app/api/db/models/user_models.py
```diff
@@ -3,45 +3,44 @@
...
 
-    roles: Optional[list["UserRole"]] = relationship(
-        "UserRole", uselist=True, back_populates="user"
-    )
+    roles: Optional[list["Role"]] = relationship("Role", back_populates="user")
 
 
-class UserRole(Base, TimestampMixin):
-    __tablename__ = "user_role"
+class Role(base.Base, base.TimestampMixin):
```
chouinar: Why change this to `base.Base` from `Base`?

app/api/db/models/user_models.py
```diff
@@ -3,45 +3,44 @@
...
-        postgresUUID(as_uuid=True), ForeignKey("user.user_id"), primary_key=True
+        postgresql.UUID(as_uuid=True), ForeignKey("user.id"), primary_key=True
     )
-    role_description: str = Column(Text, primary_key=True, nullable=False, index=True)
-
+    # Set native_enum=False to use store enum values as VARCHAR/TEXT.
+    # This is more convenient and avoids needing create a new migration
+    # when adding a new enum value.
+    # https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.Enum.params.native_enum
+    # https://medium.com/swlh/postgresql-3-ways-to-replace-enum-305861e089bc
```
chouinar: This article seems to suggest not doing it this way unless I'm misinterpreting it. Seems like it recommends native enum or foreign keys depending on the number of values

app/tests/api/db/models/factories.py
```diff
@@ -58,25 +58,25 @@ class Meta:
         sqlalchemy_session_persistence = "commit"
 
 
-class UserRoleFactory(BaseFactory):
+class RoleAssignmentFactory(BaseFactory):
```
chouinar: Should this just be `RoleFactory` so it's just `{OriginalClass}Factory`?

app/api/db/models/base.py
```diff
@@ -60,25 +56,36 @@ def copy(self, **kwargs: dict[str, Any]) -> "Base":
...
-    """Generate a tz-aware timestamp pinned to UTC"""
-    return utcnow()
+
+def same_as_created_at(context: Any) -> Any:
+    return context.get_current_parameters()["created_at"]
 
 
 @declarative_mixin
 class TimestampMixin:
+    """Mixin to add created_at and updated_at columns to a model
```
chouinar, line 74: I wonder if this mixin in particular should be a class that all others get derived from using SQLAlchemy's abstract tables (or add it to Base) as I can't any scenarios where we wouldn't want to have these timestamps on a table.

chouinar (commented):
Like the changes, just a few small questions

app/api/db/models/user_models.py
lorenyu: Not a strong reason. I think I was being flip floppy about conforming to the [google styleguide](https://google.github.io/styleguide/pyguide.html#22-imports) which was used on one of the tech challenges. The google styleguide says to only import modules not individual classes or functions.

Happy to switch it back though, since for a lot of these libraries, their documentation and examples explicitly suggest patterns that are different from the google styleguide.

app/tests/api/db/models/factories.py
lorenyu: Yeah good catch, I missed it during my renames

app/api/db/models/user_models.py
lorenyu: I'm personally fine with native, I'll remove the native=False

app/api/db/models/base.py
lorenyu: yeah i can't think of anything either but also worried about what i don't know i don't know.

what about tables that aren't used by the transactional API, like tables used in data pipelines, staging tables, and such?

app/api/db/models/user_models.py
chouinar: How does native work with regards to migrations? I think I remember not using those because alembic can't detect changes to enums so any changes require you to manually write it into alembic.

For example, use an enum and alembic should make it then alter it and run the migration create again and see what happens.

app/api/db/models/user_models.py
```diff
@@ -3,45 +3,44 @@
...
+class RoleType(str, enum.Enum):
+    USER = "USER"
+    ADMIN = "ADMIN"
 
 
-class User(Base, TimestampMixin):
+class User(base.Base, base.IdMixin, base.TimestampMixin):
     __tablename__ = "user"
 
-    user_id: Mapped[UUID] = Column(postgresUUID(as_uuid=True), primary_key=True, default=uuid_gen)
```
jamesbursa, line 26: Just my subjective preference, I like the `user_id` format as you can do `... JOIN USING (user_id)` or sometimes `NATURAL JOIN` when both tables have a `user_id`. But not sure if that would be possible with `IdMixin`.

app/api/db/models/base.py
```diff
@@ -60,25 +56,36 @@ def copy(self, **kwargs: dict[str, Any]) -> "Base":
         return copy
 
 
-def uuid_gen() -> uuid.UUID:
-    return uuid.uuid4()
+@declarative_mixin
+class IdMixin:
```
jamesbursa, line 60: Great idea to make this a mix in.

app/api/db/models/user_models.py
lorenyu: @chouinar just tried removing native=False, and I think it's less optimal from a maintenance perspective, since alembic no longer seems to generate correct migrations.

without native=false, the first migration generated is
```
sa.Column("type", sa.Enum("USER", "ADMIN", name="roletype"), nullable=False),
```
then if i run the migration and then add another value to the enum e.g.
```
class RoleType(str, enum.Enum):
    USER = "USER"
    ADMIN = "ADMIN"
    THIRD_PARTY = "THIRD_PARTY"
```
then run migrate create, it creates an empty migration that just says
```
pass
```
for both upgrade and downgrade.

on the other hand with native=False, the initial migration is
```
sa.Column(
    "type", sa.Enum("USER", "ADMIN", name="roletype", native_enum=False), nullable=False
),
```
and then the subsequent migration after adding an enum value is
```
op.alter_column(
    "role",
    "type",
    existing_type=sa.VARCHAR(length=5),
    type_=sa.Enum("USER", "ADMIN", "THIRD_PARTY", name="roletype", native_enum=False),
    existing_nullable=False,
)
```
which looks correct

lorenyu: @chouinar oh i totally missed your previous comment lol. I guess I independently confirmed what you already told me

chouinar: If I'm reading the non-native config right, it sounds like it still uses a check-constraint on the DB for the enum, just as a string? If so, then the non-native approach seems fine. Can't accidentally put bad data in. The "biggest" issue I see is that when doing DB queries there isn't a great way to know for certain all the possible values, but that's not the end of the world if you have access to the repo.

lorenyu: Actually by default there's no check constraint
image

we'd have to add `check_constraint=True` in which case i think you can see all the possible values by just doing `\d table` lemme try that out and paste here the results

lorenyu: Tried it, and here's what I found.
Initial generated migration:
image

And you can see all the possible enum values from looking in the db without looking at code by inspecting the table constraint
image

And alembic successfully detects the change in enum
image

Sounds like a good compromise to me, thoughts?

lorenyu: Actually I spoke too soon, actually running the migration to modify the enum doesn't work out of the box
```
18:29:05.580  alembic.runtime.migration            run_migrations               INFO     Running upgrade 84f3065cba80 -> 3813d6d7f022, add third party role type          
Traceback (most recent call last):
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1900, in _execute_context
    self.dialect.do_execute(
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.DuplicateObject: constraint "role_`roletype_check`" for relation "role" already exists


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/srv/api/db/migrations/run.py", line 24, in up
    command.upgrade(alembic_cfg, revision)
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/alembic/command.py", line 322, in upgrade
    script.run_env()
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/alembic/script/base.py", line 569, in run_env
    util.load_python_file(self.dir, "env.py")
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/alembic/util/pyfiles.py", line 94, in load_python_file
    module = load_module_py(module_id, path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/alembic/util/pyfiles.py", line 110, in load_module_py
    spec.loader.exec_module(module)  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/srv/api/db/migrations/env.py", line 110, in <module>
    run_migrations_online()
  File "/srv/api/db/migrations/env.py", line 104, in run_migrations_online
    context.run_migrations()
  File "<string>", line 8, in run_migrations
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/alembic/runtime/environment.py", line 853, in run_migrations
    self.get_context().run_migrations(**kw)
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/alembic/runtime/migration.py", line 623, in run_migrations
    step.migration_fn(**kw)
  File "/srv/api/db/migrations/versions/2022_12_16_add_third_party_role_type.py", line 20, in upgrade
    op.alter_column(
  File "<string>", line 8, in alter_column
  File "<string>", line 3, in alter_column
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/alembic/operations/ops.py", line 1880, in alter_column
    return operations.invoke(alt)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/alembic/operations/base.py", line 399, in invoke
    return fn(self, operation)
           ^^^^^^^^^^^^^^^^^^^
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/alembic/operations/toimpl.py", line 74, in alter_column
    operations.impl.add_constraint(constraint)
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/alembic/ddl/impl.py", line 335, in add_constraint
    self._exec(schema.AddConstraint(const))
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/alembic/ddl/impl.py", line 195, in _exec
    return conn.execute(construct, multiparams)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1380, in execute
    return meth(self, multiparams, params, _EMPTY_EXECUTION_OPTS)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/sqlalchemy/sql/ddl.py", line 80, in _execute_on_connection
    return connection._execute_ddl(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1472, in _execute_ddl
    ret = self._execute_context(
          ^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1943, in _execute_context
    self._handle_dbapi_exception(
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 2124, in _handle_dbapi_exception
    util.raise_(
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/sqlalchemy/util/compat.py", line 208, in raise_
    raise exception
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1900, in _execute_context
    self.dialect.do_execute(
  File "/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.11/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.DuplicateObject) constraint "role_`roletype_check`" for relation "role" already exists

[SQL: ALTER TABLE role ADD CONSTRAINT "role_`roletype_check`" CHECK (type IN ('USER', 'ADMIN', 'THIRDPARTY'))]
(Background on this error at: https://sqlalche.me/e/14/f405)
make: *** [db-upgrade] Error 1
```

looks like alembic isn't smart enough to drop the constraint first before creating it new

lorenyu: we could probably manually update the generated sql file to drop the constraint then add it back in, but i'm not sure that's worth it. i think SQLAlchemy already does validation at the code level, i feel like it might be more work than it's worth to add the constraint at the db level. projects can probably do that on their own if they feel the need to.

lorenyu: Pushed an update to the comment with a better explanation of why we chose native=false and constraint=false

app/api/db/models/base.py
lorenyu: actually what about linking tables for many to many relationships? it seems like we don't always need created_at / updated_at for those

app/api/db/models/user_models.py
lorenyu: That makes sense, but I feel that since the ORM will reduce the amount of SQL we'd be writing the value of that seems less than the value of what we'd get by standardizing on a primary key name, and there are also situations where the foreign key doesn't use the table name as the prefix. For example if there are two addresses `mailing_address` and `residential_address` and they both have foreign keys into the `address` table, it'd be `mailing_address_id` and `residential_address_id`, not `address_id`. I think it's still possible with `IdMixin` but it gets more complicated. I think it's possible to write an IdMixin that adds a different id name.

app/api/db/models/base.py
lorenyu: another idea could be to create a `Resource` base class, referring to what we would think of as REST API resources. all of those should have unique ids for sure. i wonder if there are other things that need ids that aren't resources.

app/api/db/models/user_models.py
lorenyu: Thinking more, i think the main argument is that most of the code we read/write will be in the application code not in SQL, so 99% of the code would be things like `user.id` vs `user.user_id` rather than what the where clause would be.

app/api/db/models/user_models.py
chouinar: Yeah, I think as-is is probably best. Enums with alembic boil down to code-only without it being a pain to manage. I think alembic lets you make your own utilities to build into it that technically it's possible to make these a proper check-constraint, but just not worth it. The only benefit is when you do need to work on raw-sql which from the projects I've been on tend to be just select queries and not inserts (outside of fixing data).

app/api/db/models/base.py
chouinar: A many-to-many table having the timestamps is potentially useful if the relationship is created at a different time than the related objects, but that's a very niche scenario for being necessary.

Staging tables that are temporary (eg load, do some operations, and delete) wouldn't have a ton of use as you'd never see them / the logs or derived tables associated with the processes would be good enough for the timestamps.

Definitely could see some sort of resource base class for this.

app/api/db/models/user_models.py
chouinar: There is an alternative that might satisfy both scenarios. If you define a column as:
`id: uuid = Column("user_id", UUID(), primary_key=True)`

Then in the DB itself, it'll create a column called `user_id`, but you use it in the code as `id`. Basically an alias. We never used that in the Mass project, but on NJ UI have used it extensively for columns that are very confusingly named. This however wouldn't let you use the mixin approach.

lorenyu: ah i see, yeah that's clever and easy enough to understand once you notice it, but if i had to guess i would think it would be slightly surprising to folks and for folks reading the code i'd guess people could easily not notice the rename. i think most people would generally expect the names of the fields to be the same as the column names.

chouinar (approved):
LGTM

PR #48, opened 2022-12-20
Create user service

zelgadis:
## Ticket

N/A

## Changes
* Created `services` directory with `users`
* Separated `create_user`, `get_user`, and `patch_user` functions into separate files
* Created `api/route/models` directory for API models
* Moved user API models into `api/route/models/user.py`

## Context for reviewers
* Part of a set of changes to separate `service` and `controller`-level concerns
* Next: user service takes API objects as inputs (eg `UserParams`), returns db objects as output (eg `User` instead of `UserResponse`)

## Testing
N/A


app/api/services/users/patch_user.py
```diff
@@ -1,73 +1,10 @@
-from datetime import date, datetime
```
zelgadis: git is being unhelpful. I deleted the `user_handler` in favor of `services/users/__init__.py`. This file just contains a lot of the code from the user_handler, so git thinks it was renamed / moved!

app/api/services/users/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .create_user import create_user  # noqa: F401
+from .get_user import get_user  # noqa: F401
+from .patch_user import patch_user  # noqa: F401
```
chouinar: If the idea is you want to export these, you can do like the DB models file and use `__all__` instead to avoid the qa issues:

https://github.com/navapbc/template-application-flask/blob/main/app/api/db/models/__init__.py

app/api/services/users/create_user.py
```diff
@@ -0,0 +1,31 @@
...
+
+from api.db.models.user_models import User, UserRole
+from api.route.api_context import ApiContext
+from api.route.models.user import UserParams, UserResponse
+
+
+def create_user(api_context: ApiContext) -> UserResponse:
+    request = UserParams.parse_obj(api_context.request_body)
+
+    user = User(
```
chouinar, line 13: I think @lorenyu will have more thoughts on this, but I feel there should be an additional persistence layer that handles the DB operations, this would make the following layers/responsibiilities:
- API route layer - handles the interaction with Flask/connexion - pretty minimal in most cases
- Service layer - business logic
- Persistence layer - DB logic

In this particular method, that would look like:
```py
def create_user(api_context: ApiContext) -> UserResponse:
    request = UserParams.parse_obj(api_context.request_body)

    # create the persistence class (I like classes so the DB session is less of a pain to pass around, could just be functions)
    db_store = DbStore(api_context.db_session)
    user_id = db_store.persist_user(request)

    for role in request.roles:
        db_store.add_role_to_user(role, user_id)

    return UserResponse.from_orm(..) # something - could just make persist_user return the user I guess
```
One other potential benefit of using a class is we could have a utility method like `get_db_store` that handles finding the proper DB session from the app so the API itself doesn't need to be aware of it via the api context. I think Loren was trying to do something like that with his DB restructuring at some point

Thoughts?

lorenyu: I told Jesse not to touch the db classes for now since i was working on that and didn't want conflicts. we could always refine the interfaces in future PRs, or he could pull in updates from main now that i merged the other PR

app/api/services/users/create_user.py
```diff
@@ -0,0 +1,31 @@
+from uuid import uuid4
+
+from api.db.models.user_models import User, UserRole
+from api.route.api_context import ApiContext
+from api.route.models.user import UserParams, UserResponse
+
+
+def create_user(api_context: ApiContext) -> UserResponse:
```
lorenyu, line 10: Todo either in this PR or a future one, but eventually the service layer shouldn't have any dependencies on the controller layer, so we'd want to slowly remove the dependency on ApiContext

lorenyu (commented):
So far looking good. Could you say a few words about the end state you're imagining?

Also FYI I'm picking up @chouinar 's old PR https://github.com/navapbc/template-application-flask/pull/20/files to see if it makes sense to migrate off of Connexion, so I might need to resolve some conflicts if we merge this. Don't worry too much about it but just letting you know.

app/api/services/users/create_user.py
zelgadis: Very much agreed! Wasn't sure to what degree we wanted to add a bunch of TODOs to the code. I suppose I could handle this by creating git issues in the future!

app/api/services/users/create_user.py
zelgadis: Very much agreed with all this! Trying to make forward progress in bite-sized PRs and not step on toes. Per Loren's other comment, I should spell out some high-level goals for this work!

app/api/services/users/__init__.py
zelgadis: oh awesome, I'll take a look!

zelgadis: Done, thanks!

app/api/services/users/create_user.py
zelgadis: I've created a git issue, and added a TODO to these!

zelgadis:
> So far looking good. Could you say a few words about the end state you're imagining?

Sure thing!

This PR:
* Separate user service logic into `services/users` module, from `route` module

Shorter term:
* Update user service module to be independent of controller concerns (eg no ApiContext, no UserResponse)
* Add tests for main user service functions (focus our testing on `services` module instead of `routes`)
* Create a README that outlines the purpose of the `routes` and `services` modules

Longer term:
* Separate db / persistence logic out from user service module into its own layer
* Create another set of REST endpoints, something like `UserData`, to further establish the above into a pattern


@lorenyu Do you want me to include the above somewhere? Maybe in a git issue?

lorenyu:
@zelgadis sure maybe create some tickets for those

lorenyu (approved):
After this, I may need your help to review a big PR. I basically picked up [where Michael left off with replacing Connexion with APIFlask](https://github.com/navapbc/template-application-flask/pull/20/files), and [I have a branch](https://github.com/navapbc/template-application-flask/compare/lorenyu/apiflask?expand=1) that I think gets there. But I'll need to resolve merge conflicts first after you merge this.

jamesbursa:
Nit - what are your thoughts on `service` / `user` instead of plural? IMHO using singular matches the contents, makes the import read better (`import service.user`) and avoids dealing with the different ways that English does plural (addresses, children, taxes, etc.)

lorenyu:
@jamesbursa what about `services/user`? I think `user` makes sense, but I think pluralizing `services` makes sense since it's more of a directory of services. We can do `from services import user` or `import services.user`.

Other options: `services/user_service` or `services/user_services`?

PR #51, opened 2022-12-20
Replace connexion/pydantic with apiflask/marshmallow

### LINKED ISSUE #12: Look into alternatives to connexion for flask openAPI support

lorenyu:
## Ticket

Resolves #12 

## Changes
Replace connexion/pydantic with apiflask/marshmallow

## Context for reviewers
Based on our experience we decided to switch off of Connexion and do a code-first approach to openAPI spec, and [decided on APIFlask](https://github.com/navapbc/template-application-flask/blob/main/docs/decisions/0001-connexion-replacement.md).

This PR rips out connexion and uses APIFlask instead

APIFlask uses marshmallow instead of pydantic, so we may want to consider moving off of Pydantic fully in a future ticket so we don't have two different runtime type checkers.

## Testing
- Ran server and loaded healthcheck
- Ran make check


lorenyu:
@zelgadis @chouinar heads up, i need to resolve merge conflicts but will need your help reviewing this afterwards

app/api/services/users/create_user.py
```diff
@@ -1,27 +1,20 @@
...
-from api.route.models.user import UserParams, UserResponse
 
 
 # TODO: separate controller and service concerns
 # https://github.com/navapbc/template-application-flask/issues/49#issue-1505008251
-def create_user(api_context: ApiContext) -> UserResponse:
-    request = UserParams.parse_obj(api_context.request_body)
+def create_user(user_data: dict, api_context: ApiContext) -> User:
+    # TODO: move this code to service and/or persistence layer
+    user = User(**user_data)
```
chouinar, line 13: Before I dig into this too much, are you planning to adjust this? I'd really like to not have dictionaries. In this case it's fine because the request model matches the DB model, but that won't always be the case for every API, how do we process that into a meaningful object? Marshmallow-dataclasses helped this, but then we hit the `partial=true` patch issue.

lorenyu: if the names don't match i was thinking we'd use [serialization/deserialization keys](https://marshmallow.readthedocs.io/en/stable/quickstart.html#specifying-serialization-deserialization-keys) in the schema to do the mapping/renaming of keys

lorenyu: or is there a way to keep it as a dict for patch but use [post_load](https://marshmallow.readthedocs.io/en/stable/quickstart.html#deserializing-to-objects) for the other endpoints to convert to an object?

app/tests/conftest.py
```diff
@@ -172,7 +172,7 @@ def app(test_db_session):
 
 @pytest.fixture
 def client(app):
-    return app.app.test_client()
+    return app.test_client()
```
zelgadis, line 175: 👍

app/api/services/users/create_user.py
chouinar: I think any coupling of request to the DB schema should be seen as an ideal scenario, but not a required one. In this case where we're making up our own endpoint it's simple to make the request and DB models match, but that doesn't always hold up (eg. NJ where we didn't get to decide on how either was structured and 1 request maps to 20 tables in a different shape with 100 calculated params in-between).

Let's assume this fairly simple scenario. You have this POST request:
```
{
   "name": "bob",
   "age": 20,
   "favorite_colors": ["blue", "red"],
}
```
Now lets assume that the DB has a single table with the following columns (I know you wouldn't structure it this way):
- name: str
- age: int
- likes_blue: bool
- likes_red: bool

Ideally I'd be able to work with creating that model by doing something like:
```py
db_record = NameObject(
    name=request.name,
    age=request.age,
    likes_blue="blue" in request.favorite_colors,
    likes_red="red" in request.favorite_colors,
)
```
If request were a dictionary there would be a bit of issue with typing I'd imagine (I'm actually surprised it's okay with the patch endpoint approach we have as-is). And beyond the typing issue, it would be more error-prone/effort for development to have to look back at the model to figure out how the dict it validated is structured.

---

Serializing it into an object is good, but besides marshmallow dataclasses, is there any way to do that without re-declaring the model again (which might be fine, plus pydantic is convenient for loading from JSON in that case).

zelgadis: I don't have an answer to the question, but +1 to avoiding using dicts if possible

app/api/services/users/create_user.py
```diff
@@ -1,27 +1,20 @@
...
-    if request.roles is not None:
-        user.roles = [Role(user_id=user.id, type=role.type) for role in request.roles]
+    if user.roles is not None:
+        for role in user.roles:
+            role.user_id = user.id
 
-    return UserResponse.from_orm(user)
+    api_context.db_session.add(user)
+    api_context.db_session.flush()
+    return user
```
zelgadis, line 22: 👍

app/api/services/users/get_user.py
```diff
@@ -1,12 +1,17 @@
...
 # TODO: separate controller and service concerns
 # https://github.com/navapbc/template-application-flask/issues/49#issue-1505008251
-def get_user(user_id: str, api_context: ApiContext) -> UserResponse:
-    user = get_or_404(api_context.db_session, User, user_id)
+def get_user(user_id: str, api_context: ApiContext) -> User:
+    # TODO: move this to service and/or persistence layer
+    result = api_context.db_session.query(User).options(orm.selectinload(User.roles)).get(user_id)
+
+    if result is None:
+        raise apiflask.HTTPError(404, message=f"Could not find user with ID {user_id}")
```
zelgadis: Can we put this in `user_route.py::user_get`? I think it's reasonable for this service method to return `None`, and the controller can re-package that as a 404 for the end user.

app/api/services/users/patch_user.py
```diff
@@ -26,11 +24,11 @@ def patch_user(user_id: str, api_context: ApiContext) -> UserResponse:
     api_context.db_session.flush()
     api_context.db_session.refresh(user)
 
-    return UserResponse.from_orm(user)
+    return user
 
 
 def _handle_role_patch(
-    user: User, request_role_types: Optional[list[RoleType]], api_context: ApiContext
+    user: User, request_roles: Optional[list[Role]], api_context: ApiContext
```
zelgadis, line 33: 👍

app/tests/api/route/test_user_route.py
```diff
@@ -250,6 +233,8 @@ def test_patch_user_200_roles(
     # Add two roles
     request = {"roles": [{"type": "ADMIN"}, {"type": "USER"}]}
     response = client.patch(f"/v1/user/{user.id}", json=request, headers={"X-Auth": api_auth_token})
+    print(response.get_json())
```
zelgadis: oops!

app/api/route/response.py
```diff
@@ -28,54 +28,21 @@ def __init__(
...
-    code = status_code.code if status_code.code is not None else 400
-    return Response(status_code=code, message=message, errors=errors, data=data, warnings=warnings)
+    message: str
+    data: Optional[Base] = None
+    warnings: list[ValidationErrorDetail] = dataclasses.field(default_factory=list)
+    errors: list[ValidationErrorDetail] = dataclasses.field(default_factory=list)
+
+    # This method is used to convert ApiResponse objects to a dictionary
+    # This is necessary because APIFlask has a bug that causes an exception to be
+    # thrown when returning objects from routes when when BASE_RESPONSE_SCHEMA is set
```
zelgadis:
```suggestion
    # thrown when returning objects from routes when BASE_RESPONSE_SCHEMA is set
```

app/api/route/schemas/user_schemas.py
```diff
@@ -0,0 +1,53 @@
...
+# Role Models
+##############
+
+
+class RoleSchema(marshmallow.Schema):
+    type = marshmallow_fields.Enum(RoleType, description="The name of the role", by_value=True)
+
+    # Output only fields
+    created_at = fields.DateTime(dump_only=True)
+    updated_at = fields.DateTime(dump_only=True)
```
zelgadis, line 19: 👍

app/api/route/schemas/user_schemas.py
```diff
@@ -0,0 +1,53 @@
...
+        required=True,
+    )
+    is_active = fields.Boolean(
+        description="Whether the user is active",
+        required=True,
+    )
+    roles = fields.List(fields.Nested(RoleSchema), required=True)
+
+    created_at = fields.DateTime(dump_only=True)
+    updated_at = fields.DateTime(dump_only=True)
```
zelgadis, line 54: Should we annotate these as output-only, like the fields in `Role`?

app/api/route/schemas/user_schemas.py
```diff
@@ -0,0 +1,53 @@
...
+        return Role(**data)
+
+
+##############
+# User Models
+##############
+
+
+class UserSchema(marshmallow.Schema):
+    user_id = fields.UUID(dump_only=True)
```
zelgadis: Hmmm, shouldn't this just be `id`? 🤔

app/api/route/schemas/user_schemas.py
```diff
@@ -0,0 +1,53 @@
...
+        pattern=r"^([0-9]|\*){3}\-([0-9]|\*){3}\-[0-9]{4}$",
+    )
+    date_of_birth = fields.Date(
+        description="The users date of birth",
+        required=True,
+    )
+    is_active = fields.Boolean(
+        description="Whether the user is active",
+        required=True,
+    )
```
zelgadis, line 49: not really a comment on this PR, but it seems strange to make `is_active` a required field to be set through the API. I would assume we could make this optional and default it to `true` for new users!

zelgadis (commented):
Looking good, thanks for making these changes! I have a few small requested changes and comments. Happy to re-review once merge conflicts are fixed / the build is passing!

app/api/services/users/get_user.py
lorenyu: good idea, i wasn't thinking about the service/controller/persistence separation in this PR, just the replace connexion with apiflask work. could i add a TODO comment for that or do you think it should be part of this PR?

app/api/route/schemas/user_schemas.py
lorenyu: they are marked as `dump_only=True`. do you mean something else?

app/api/route/schemas/user_schemas.py
lorenyu: oh funny that the tests didn't catch that, i'll add an assertion

app/api/route/schemas/user_schemas.py
lorenyu: Yeah I also am wondering whether we should come up with other examples that might be more realistic, since I think it's also not common on our projects to have the client set the role. But let's do that for another PR

app/api/services/users/create_user.py
lorenyu: with that example, the request schema would be something like

```
class PersonSchema(marshmallow.Schema):
  name = fields.String()
  age = fields.Integer()
  favorite_colors = fields.List(fields.String) # could be fields.Enum instead of fields.String but that's not relevant to our discussion
```

and then in # db.models.person we'd have 

```
class Person(Base):
  name = Column(Text)
  likes_blue = Column(Boolean)
  likes_red = Column(Boolean)
```

then in our controller we'd do
```
def create_person(person_request_data: dict):
  person = Person(
    name=person_request_data["name"],
    age=person_request_data["age"],
    likes_blue="blue" in person_request_data["favorite_colors"],
    likes_red="red" in person_request_data["favorite_colors"],
  )
```

and we'd still get runtime API request validation already from the Marshmallow schema, and we'd still get runtime db model constructor level validation with the SQLAlchemy model constructors.

@zelgadis yeah I started out having objects, using a marshmallow schema post_load hook like

```
class PersonSchema(Schema):
  ...
  @post_load
  def make_person(self, data: dict):
    return Person(
      name=data["name"],
      age=data["age"],
      likes_blue="blue" in data["favorite_colors"],
      likes_red="red" in data["favorite_colors"],
    )
```
but the issue is with the patch endpoint, where we want to allow clients to pass requests that don't have fields that are marked as required in the schema, and where we want to know exactly what fields were passed and not default them to None, since there's a semantic difference between PATCH set field X to null vs PATCH request that doesn't have field X. but if we convert it to an object in that case, the object will be an instance of the class, which means it will have all the attributes.

as such i think dictionaries with runtime type validation seems like the best approach.

lorenyu: As a thought exercise, if i were to ask you to define a class `UserPatchParams` that represents the request data for `PATCH /v1/users/:user_id`, how would you define it?

lorenyu: As a point of comparison, [this](https://github.com/navapbc/template-application-flask/blob/cdc883299e518d1c295c3b472c1e31a739bf690e/app/api/route/request.py) is what we needed to do in order to support PATCH requests with pydantic models. What it's doing is that a pydantic class will keep track of the fields that were set in a hidden field `__fields_set__`, and we use that to reverse engineer the dict that was originally used to create the instance of the pydantic model, but at that point I don't see how that's different from using a `dict` directly and just cutting out the middle step of turning it into an object first (in this PR that file isn't needed anymore).

Taking a step back, here's the lifetime of the data
HTTP request (byte string) --> Flask framework (parse json to dict) --> App controller / route handler (runtime validation of request data) --> App service (work with domain objects)

And what we're discussing is whether we should add an extra layer of "convert request data to request data object" at the controller level.

lorenyu:
Looks like I need to get this PR – https://github.com/navapbc/template-application-flask/pull/41/files – merged before the lint-security check will pass, so if anyone could do a quick review of that that'd be great too

app/api/route/schemas/user_schemas.py
zelgadis: Oh I meant like how we have an `Output only fields` comment on the similar fields in the Role model. Was just curious why we had that comment in one place and not the other!

https://github.com/navapbc/template-application-flask/pull/51/files/1e8d62a65a2018b19daf883bb23af9ecc679a2e2#diff-2562ba6ed34409e97217de0461a40192fc0723bd309659f1a18ea0b47546a7ddR17-R19

But I see that we have other fields that are dump_only on User, so 🤷. I'm gonna mark this as resolved!

app/api/services/users/get_user.py
zelgadis: No strong feelings!

zelgadis:
> Looks like I need to get this PR – https://github.com/navapbc/template-application-flask/pull/41/files – merged before the lint-security check will pass, so if anyone could do a quick review of that that'd be great too

Done!

app/api/services/users/create_user.py
zelgadis: @lorenyu Okay gotcha, thanks for explaining! I see what value the dict is adding here and I would be okay with this approach. I still think I prefer the old approach of passing request data objects (eg `UserPatchParams`) to the service methods, and enumerating the `get_set_fields` in the PATCH case!

zelgadis: > As a thought exercise, if i were to ask you to define a class `UserPatchParams` that represents the request data for `PATCH /v1/users/:user_id`, how would you define it?

This got my thinking cap on! I don't have a good sense of whether or not what I'm about to propose is a good idea or not, but I'm going to share anyways 😛 

In the `UserPatchParams` paradigm, we have fields marked with `Optional`, since they don't need to be specified through the API:

```
first_name: Optional[str]
```

This is essentially the same as:

```
first_name: Union[str, None]
```

I wonder if we could extend that to include an `Omitted` option as well? (`Omitted` would just be an empty class)

```
first_name: Union[str, None, Omitted]
```

In the constructor for the `UserPatchParams`, all fields could be `Omitted` by default, and only set to a value or `None` if they were included in the request. Then you could loop over the fields and set them only if they were not `Omitted`:

```
for attribute, value in params.__dict__.items():
  if type(value) == Omitted:
    # this attribute was not set in the request - skip it!
    continue

  setattr(user, attribute, value)
```

This _feels_ un-pythonic and perhaps over-engineered to me 🤔🤔🤔 But what it does that I like:
* Allows us to pass data objects instead of dicts to the service methods, which I think will make them cleaner and allow for easier testing
* No need for a separate method call or hidden data store to figure out which attributes were set by the caller - the model can hold this information itself. What you see is what you get. If an attribute is `None`, that means it was set to `None`.

chouinar: I did have something in an earlier version of this to handle the "was it set in the request" problem. Basically, it re-references the raw-json to see if the value was present or not. Don't think it was robust enough for the heavily-nested objects, but worked in the simple case.

An idea that might overcomplicate things, but not sure if it'll work at all. Would it be possible to define the models in Pydantic, so we can use them like the existing approach, but export the pydantic schema (see: https://docs.pydantic.dev/usage/schema/) and somehow import that into marshmallow schema (see: https://marshmallow.readthedocs.io/en/stable/quickstart.html#creating-schemas-from-dictionaries ) - Would probably require some conversion utils, but maybe? Although for that, it'd just be easier to switch entirely to fastapi which is literally pydantic->openapi - https://fastapi.tiangolo.com/

app/api/route/schemas/user_schemas.py
lorenyu: oh i see what you mean. i like keeping `id` at the top and it's output only. what do you think about a comment like
```
# Output only fields in addition to id field
```

app/api/services/users/get_user.py
lorenyu: I'll put a TODO, thanks!

app/api/services/users/create_user.py
lorenyu: @zelgadis  Yeah, the having the Union with an Omitted value could work, and I agree with easier testing, but also agree that it feels overengineered, and I really want something simple that someone can look at and understand easily.

@chouinar using pydantic and marshmallow together could work but also feels a bit heavy handed.

I'm open to using fastapi at some point, but worried about scope creep. Also [Flask dwarfs all other frameworks in terms of downloads/day](https://piptrends.com/compare/flask-vs-fastapi-vs-django-vs-pyramid), so it feels like most people are able to make it work and be satisfied with it ... 🤔 

why is this so hard...

lorenyu: @zelgadis @chouinar I also played around with the approach of defining Params classes, and it gets really complicated really fast with nested objects

There's only one idea that I can think of kind that I think is a reasonable approach:
* Use the Marshmallow Schemas in the apiflask route decorators to do runtime validation of API request bodies coming from client requests
* **Also** do runtime validation of types within service methods. We can reuse the Marshmallow Schemas too.

e.g.
```
### user_route.py (controller layer)
@user_blueprint.patch("/v1/user/<uuid:user_id>")
@user_blueprint.input(user_schemas.UserSchema(partial=True))
@user_blueprint.output(user_schemas.UserSchema)
@user_blueprint.auth_required(api_key_auth)
def user_patch(user_id: str, user_patch_params: dict) -> dict:
    # user_patch_params has already been validated to match UserSchema based on apiflask framework
    user = user_service.update_user(user_id, user_patch_params)
    return response.ApiResponse(message="Success", data=user).asdict()

------------------------

### user_service/update_user.py (service layer)
def update_user(user_id: uuid.UUID, fields: dict):
    # validate input using marshmallow schema
    fields = user_schemas.UserSchema().load(fields)
    ...
```

**Pros:**
* The advantage of this approach is the service code is now more self-contained, a developer reading the code can see that the service is doing run-time validation of the input and feel confident that it works without looking at the controller layer.
* Furthermore, this adds runtime type safety to the service methods in a self-contained fashion, even if it's called outside of an API request standpoint (e.g. if it's called from a batch job or CLI). The fact that the data is validated twice in the API request context shouldn't really be an issue. The performance impact is negligible.

What do y'all think? I still think this is simpler to understand and requires a lot less magic than what we were doing with Connexion. I feel good about runtime type safety and not trying to force static type safety ultimately the data is coming from the internet at runtime as a JSON string, so we can't get around the fact that the type safety can only be achieved at runtime.


Separately, an idea to help make testing easier: what if we made a Faker provider that takes a marshmallow schema and uses that to generate data. We could use it to generate dictionaries that map to dictionaries that would be received by the API.

chouinar: What type is the output of the `load` function? If it's still a dict, that doesn't help. My concern is that doing development with a dict is incredibly painful, I have to constantly refer back to the object manually and none of my IDE functionality works (auto-complete, type hints, and so on). In addition, `val = fields["some_param"]` is always going to run into typing issues.

If I started with a system that did it this way, the first thing I'd do is go create a pydantic model to load that JSON into as keeping that in sync with the request model is going to save me several magnitudes of dev-time compared to leaving it a raw dict.

lorenyu: `load` returns a `dict`, but the amount of time it will "live" as a `dict` is basically nothing, since the service will immediately construct a domain object immediately from the request params i.e. either
```
def update_user(data):
  # validate data
  user = User(first_name=data["first_name"]...)
```
or
```
  user = User(**data)
```
so that's basically equivalent to converting it to a Pydantic model. but it removes the need to have another entire layer of classes in between.

So in my previous suggestion, the life cycle is "network json string" --> "dict" --> validate dict --> convert to domain object/ and in your suggestion it's adding an additional step in between so it's like "network json string" --> "dict" --> validate dict --> convert to request object --> convert to domain object. But my argument is that "convert to request object" isn't a valuable step since the service is immediately going to "convert to domain object" as the first first step of its operation.

chouinar: In this particular endpoint I agree, any additional conversion is basically redundant because the request model is the same as the DB model. But this is the happy-path case, this is how we'd love to setup most endpoints, but what about the complex scenario?

For example, the update application endpoint in Mass PFML. The `ApplicationRequestBody` needed a lot of parsing and pulling out of values in order to update various tables. Doing that with a dictionary would be quite painful / error prone.

It might be that the best approach if they were using this API structure would be to make a Pydantic model for this specific endpoint, but for other simpler endpoints skip that as it's not worthwhile. Does that seem reasonable? If so, could we add an example endpoint that does that to help show how you might do that?

zelgadis: It seems like we're having difficulty reaching consensus on this. I tend to agree with @chouinar but I don't want to block forward progress.. Would it be helpful to spend 30 - 45 min discussing this live?

Some questions I have:
* What are we trying to gain by moving away from the `get_set_parameters` approach?
* Why does making Params classes for nested objects get complicated? Is it just that we have to create a Params class for each nested object?
* What is the "connexion magic" that we're trying to avoid?

zelgadis (approved):
LGTM, thanks for doing this work! I have a preference for eliminating our use of dicts in service methods, but I'm fine revisiting that later. Would love to see some TODOs added in where we're using the dicts!

PR #53, opened 2022-12-22
Add application level Python audit logging

jamesbursa:
## Ticket

None

## Changes

- Add module `api.logging.audit` that implements Python audit logging.
- Add module `api.util.collections.dict` with `class LeastRecentlyUsedDict`.

## Context for reviewers

Python has an audit events feature. This code uses the feature to log audit events, which can be useful for security analysis.

The logs have a custom logging level `AUDIT`.

## Testing

A selection of logs when the server starts, include file opens, socket operations, and subprocess start:

Screenshot_2022-12-22_14-42-54


app/api/logging/audit.py
```diff
@@ -0,0 +1,71 @@
...
+    "sys._getframe",
+    "sys.settrace",  # interferes with PyCharm debugger
+}
+
+
+def audit_hook(event_name: str, args: tuple[Any, ...]) -> None:
+    if event_name in IGNORE_AUDIT_EVENTS:
+        return
+    if event_name == "open" and isinstance(args[0], str) and "/__pycache__/" in args[0]:
+        return
```
zelgadis: nit: Can we add a comment about this special case?

app/api/logging/audit.py
```diff
@@ -0,0 +1,71 @@
...
+    if event_name == "os.chmod" and type(args[0]) is int:
+        # Gunicorn generates a high volume of these events in normal operation (see workertmp.py)
+        return
+    # Disable audit log for timezone files.
```
zelgadis: nit: I generally prefer to put comments inside the if block indent!

```
if event_name == "open" and isinstance(args[0], str) and "/pytz/" in args[0]:
    # Disable audit log for timezone files.
    return
```

zelgadis: also, I'd appreciate an explanation in the comment as to _why_ we're ignoring timezone files!

app/api/util/collections/dict.py
```diff
@@ -0,0 +1,36 @@
+#
+# Dictionaries with additional behaviour.
+#
+
+import collections
+
+
+class LeastRecentlyUsedDict(collections.OrderedDict):
```
zelgadis, line 9: should this file be named `least_recently_used_dict.py`?

zelgadis: And in `app/api/util/collections/__init__.py` we can do:

```
from .least_recently_used_dict import LeastRecentlyUsedDict
```

And to use it:

```
from api.util.collections import LeastRecentlyUsedDict

LeastRecentlyUsedDict(maxsize=4)
```

Which I think is a little cleaner than:

```
from api.util.collections.dict import dict_util

dict_util.LeastRecentlyUsedDict(maxsize=4)
```

zelgadis (approved):
LGTM! Made a few suggestions / nits, but nothing blocking 👍

app/api/util/collections/dict.py
jamesbursa: Used the name dict as this could contain other dict related utilities in future. I added the import to `__init__.py`.

Issue #52, opened 2022-12-22
Use classes / objects as inputs to service methods


zelgadis:
Per discussion here: https://github.com/navapbc/template-application-flask/pull/51#discussion_r1053724252
and here: https://nava.slack.com/archives/C03G1SWD9H7/p1671033815911879

We want to avoid using dicts in our service methods.

PR #56, opened 2022-12-28
Pass objects instead of dicts to service methods

zelgadis:
## Ticket

Issue #52 

## Changes
* Add `UserPatchParamsSchema` and `UserPostParamsSchema`, which both apply validations from `UserSchema`
* Pass `UserPatchParams` to `user_patch`
* Pass `User` to `user_create`
* Add PatchParamsBase to handle persisting original JSON input in a PATCH request
* Added tests for various `user_patch` scenarios

## Context for reviewers
Based on discussion [here](https://github.com/navapbc/template-application-flask/pull/51#discussion_r1053724252) and [here](https://nava.slack.com/archives/C03G1SWD9H7/p1671033815911879).

The meat of the change comes from `UserPatchParamsSchema` and `UserPostParamsSchema`. Each enforces validations from the inherited `UserSchema`, but produces different objects via the `post_load` function. This does a couple things:
* Input is validated against the same `UserSchema` regardless of if the `POST` or `PATCH` endpoints are being hit
* validation rules do not need to be redefined, since the schemas inherit from `UserSchema`
* The `UserPatchParamsSchema`'s `post_load` function creates a `UserPatchParams` object, that persists the original JSON payload, for use in `user_patch`

## Testing
N/A

### DIFF
--- a/app/api/route/schemas/user_schemas.py
+++ b/app/api/route/schemas/user_schemas.py
@@ -5,7 +5,7 @@ from apiflask import fields
 from marshmallow import fields as marshmallow_fields
 
 from api.db.models.user_models import Role, RoleType, User
-from api.models.user import UserPatchParams
+from api.services.models.user import UserPatchParams
 
 ##############
 # Role Models
--- a/app/api/route/schemas/user_schemas.py
+++ b/app/api/route/schemas/user_schemas.py
@@ -4,7 +4,7 @@ import marshmallow
 from apiflask import fields
 from marshmallow import fields as marshmallow_fields
 
-from api.db.models.user_models import Role, RoleType
+from api.db.models.user_models import Role, RoleType, User
 
 ##############
 # Role Models
@@ -52,3 +52,7 @@ class UserSchema(marshmallow.Schema):
     # Output only fields in addition to id field
     created_at = fields.DateTime(dump_only=True)
     updated_at = fields.DateTime(dump_only=True)
+
+    @marshmallow.post_load
+    def make_user(self, data: dict, **kwargs: dict[str, Any]) -> User:
+        return User.create_from_data(data)
--- a/app/api/route/schemas/user_schemas.py
+++ b/app/api/route/schemas/user_schemas.py
@@ -66,4 +66,8 @@ class UserPostParamsSchema(UserSchema):
 class UserPatchParamsSchema(UserSchema):
     @marshmallow.post_load
     def make_user_patch_params(self, data: dict, **kwargs: dict[str, Any]) -> UserPatchParams:
-        return UserPatchParams(data)
+        print("IN make_user_patch_params")
+        params = UserPatchParams(**data)
+        params.load_stuff(data)
+
+        return params
--- a/app/api/route/schemas/user_schemas.py
+++ b/app/api/route/schemas/user_schemas.py
@@ -66,5 +66,4 @@ class UserPostParamsSchema(UserSchema):
 class UserPatchParamsSchema(UserSchema):
     @marshmallow.post_load
     def make_user_patch_params(self, data: dict, **kwargs: dict[str, Any]) -> UserPatchParams:
-        print("IN make_user_patch_params")
         return UserPatchParams(**data)
--- a/app/api/route/schemas/user_schemas.py
+++ b/app/api/route/schemas/user_schemas.py
@@ -54,10 +54,6 @@ class UserSchema(marshmallow.Schema):
     created_at = fields.DateTime(dump_only=True)
     updated_at = fields.DateTime(dump_only=True)
 
-    @marshmallow.post_load
-    def make_user(self, data: dict, **kwargs: dict[str, Any]) -> User:
-        return User.create_from_data(data)
-
 
 # TODO: dataclass?
 # TODO: maybe extend User?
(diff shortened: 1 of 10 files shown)


app/api/route/schemas/user_schemas.py
```diff
@@ -52,3 +53,19 @@ class UserSchema(marshmallow.Schema):
...
     created_at = fields.DateTime(dump_only=True)
     updated_at = fields.DateTime(dump_only=True)
+
+
+# Validate input according to the UserSchema
+# and convert input into User
+class UserPostParamsSchema(UserSchema):
+    @marshmallow.post_load
+    def make_user_post_params(self, data: dict, **kwargs: dict[str, Any]) -> User:
+        return User(**data)
```
zelgadis, line 63: It might make more sense to return a `UserPostParams` object here instead of a `User`. Thoughts? I kept it this way for now so that I could keep changes as minimal as possible, and not have to change much in `create_user`.

app/api/services/users/patch_user.py
```diff
@@ -3,15 +3,15 @@
...
 
 # TODO: separate controller and service concerns
 # https://github.com/navapbc/template-application-flask/issues/49#issue-1505008251
-# TODO: Use classes / objects as inputs to service methods
-# https://github.com/navapbc/template-application-flask/issues/52
-def patch_user(user_id: str, patch_data: dict, api_context: ApiContext) -> User:
+def patch_user(user_id: str, params: UserPatchParams, api_context: ApiContext) -> User:
     user = get_or_404(api_context.db_session, User, user_id)
 
+    patch_data = params.get_set_params()
```
zelgadis, line 14: Re-introducing this pattern as the solution for figuring out what attributes were set by the user.

app/api/services/models/user.py
```diff
@@ -0,0 +1,6 @@
+from .base import PatchParamsBase
+
+
+# Inherit from PatchParamsBase, to persist the original JSON input
+class UserPatchParams(PatchParamsBase):
```
zelgadis, line 5: I decided not to add any fields to the `UserPatchParams` class, since we ignore them and just use the input JSON dict instead.

app/api/route/schemas/user_schemas.py
lorenyu: yeah i think that makes sense to me!

app/api/route/schemas/user_schemas.py
```diff
@@ -4,7 +4,8 @@
 from apiflask import fields
 from marshmallow import fields as marshmallow_fields
 
-from api.db.models.user_models import Role, RoleType
+from api.db.models.user_models import Role, RoleType, User
+from api.services.models.user import UserPatchParams
```
lorenyu, line 8: good call putting `UserPatchParams` in the services package. i think that makes more sense than in the `route` folder like i had it in [my PR](https://github.com/navapbc/template-application-flask/pull/57/files) since the services package shouldn't depend on anything specific to the web/api context.

on another note, i suggest we do
```
from api.services.models import user
```
and then use `user.UserPatchParams` so we don't accidentally re-export `UserPatchParams` from this module (see my other comment about that)

app/api/services/users/patch_user.py
```diff
@@ -3,15 +3,15 @@
 from api.db.models.user_models import Role, User
 from api.route.api_context import ApiContext
 from api.route.route_utils import get_or_404
+from api.route.schemas.user_schemas import UserPatchParams
```
lorenyu, line 6: i noticed that we import `UserPatchParams` from `services` in the other file, but from `route` in this file. should we import it from `services`? This situation makes me realize an additional value of [google's python style guide on imports](https://google.github.io/styleguide/pyguide.html#22-imports), which suggests in most cases to do

```
from api.services import models
# use models.UserPatchParams
```

rather than

```
from api.services.models UserPatchParams
```

i think it helps avoid accidentally re-exporting classes i.e. i don't think it was intentional for the `user_schemas` module to be re-exporting `UserPatchParams`

app/api/route/schemas/user_schemas.py
chouinar: Not necessary for this endpoint, but how would you implement this if the POST body didn't match the structure of the DB model? I think we'd want an intermediate like the UserPostParams you're suggesting, that way we can also keep the DB model creation/updates down in the persistence layer.

app/api/services/models/base.py
```diff
@@ -0,0 +1,12 @@
+# Base class for handling input params for PATCH endpoints.
+# Persist the original JSON input in _fields_set, so that
+# we can determine which fields were set by the user, and
+# which were omitted.
+class PatchParamsBase:
```
lorenyu, line 5: I really like the intent to generalize and reuse the idea of PatchParams and get_set_params. In this case I don't think we need inheritance since we don't need polymorphism i.e. the functionality of `get_set_params` should be the same in all cases, and the only difference is in the "ApiModel" (in our case it's the "ApiUserModel").


I suggest we do something like

```
RequestModelType = TypeVar("RequestModelType", bound=BaseParams)
@dataclasses.dataclass
class PatchParams:
    params: RequestModelType
    fields_set: list[str]
```

then to use it 

```
class UserPatchParamsSchema(UserSchema):
    @marshmallow.post_load
    def make_user_patch_params(self, data: dict, **kwargs: dict[str, Any]) -> UserPatchParams:
        return PatchParams(
            params=UserParams(**kwargs),
            fields_set=list(kwargs.keys()), # just storing the keys feels less redundant to me
        )
```

lorenyu (commented):
left some thoughts! lemme know your thoughts too. also, what do you think about whether we should continue from your branch or from mine?

zelgadis:
Closing this, since this work was addressed in @lorenyu 's https://github.com/navapbc/template-application-flask/pull/57!

PR #57, opened 2022-12-28
Use TypedDict for type safety of request data

### LINKED ISSUE #52: Use classes / objects as inputs to service methods

### LINKED ISSUE #58: Generated openapi spec field order is non-deterministic

### LINKED ISSUE #61: Add utility method to PatchParams for iterating over the set fields from the params object
see https://github.com/navapbc/template-application-flask/pull/57#discussion_r1060937667

lorenyu:
## Ticket

Resolves #52 
Resolves #61 
Resolves #60 

## Changes
* Add generic PatchParams class
* Define CreateUserParams and PatchUserParams TypedDict types to represent
  parameters to create_user and patch_user service methods
* Remove now unused module api.route.models
* Fix issue in apiflask where generated openapi schema fields had nondeterministic order

## Context for reviewers
This came out of the todos that were left from PR #51. @zelgadis you and I both tackled this (his PR is #56) and we ended up with a similar approach. I incorporated some of the changes from your PR into this PR, but happy to go with your PR or mine. This PR also fixes a bug in the openapi.yaml generation, where the order of fields isn't deterministic – in a future PR we should just be able to update apiflask and then we can remove the workaround.

## Testing
CI


app/api/route/healthcheck.py
```diff
@@ -1,18 +1,18 @@
...
 import api.logging
 from api.route import response
 from api.route.api_context import api_context_manager
+from api.route.schemas import request_schema
 
 logger = api.logging.get_logger(__name__)
 
 
-class HealthcheckSchema(marshmallow.Schema):
+class HealthcheckSchema(request_schema.OrderedSchema):
```
lorenyu, line 15: see OrderedSchema below in request_schema.py, this was needed to fix an issue where generated openapi schemas weren't deterministic

lorenyu:
I pulled out the openapi order fix into a separate PR so that this PR can be easier to review once that's merged in: https://github.com/navapbc/template-application-flask/pull/60

app/api/services/users/models.py
```diff
@@ -0,0 +1,24 @@
...
+
+@dataclasses.dataclass
+class CreateUserParams:
+    first_name: str | None = None
+    middle_name: str | None = None
+    last_name: str | None = None
+    phone_number: str | None = None
+    date_of_birth: date | None = None
+    is_active: bool | None = None
+    roles: list[RoleParams] | None = None
```
chouinar: Hmm, this does make working with the object have one annoyance. Let's assume `first_name` is always required based on the schema. We as developers know this because the validation happened, but in-code we'd have to check that it's not None to make mypy happy. `cast` helps, but it can't trace that through the code very well so anywhere you use it and it being null is a problem, mypy will require you to cast again.

This might be wishful thinking, but it'd be great if you could define a dataclass as:
```
@dataclass
class MyDataclass:
     param1: str
     param2: str | None
     param3: str
```
and then generate another class with identical params, but all of them optional. It might be possible to do that via https://docs.python.org/3/library/dataclasses.html#dataclasses.make_dataclass by iterating over the original classes fields, but is that making things too abstract?

app/api/services/users/models.py
```diff
@@ -0,0 +1,24 @@
...
+from api.services.core import patch_params
+
+
+@dataclasses.dataclass
+class RoleParams:
+    type: user_models.RoleType
+
+
+@dataclasses.dataclass
+class CreateUserParams:
```
chouinar: I get organizing these service objects into a separate file, but also find it a bit annoying having to jump back and forth, not sure if there's a better way, any thoughts?

app/api/route/schemas/user_schemas.py
```diff
@@ -52,3 +59,18 @@ class UserSchema(marshmallow.Schema):
...
+
+class CreateUserSchema(UserSchema):
+    @marshmallow.post_load
+    def make_user(self, data: dict, **kwargs: dict) -> users.CreateUserParams:
+        return users.CreateUserParams(**data)
+
+
+class PatchUserSchema(UserSchema):
+    @marshmallow.post_load
+    def make_user(self, data: Mapping[str, Any], **kwargs: dict) -> users.PatchUserParams:
```
chouinar: You can do `dict[str, Any]` as of 3.10

app/api/route/schemas/user_schemas.py
```diff
@@ -1,34 +1,41 @@
...
+from typing import Any, Mapping
 
 import marshmallow
 from apiflask import fields
 from marshmallow import fields as marshmallow_fields
 
-from api.db.models.user_models import Role, RoleType
+from api.db.models import user_models
+from api.route.schemas import request_schema
+from api.services import users
```
chouinar: I'm curious if you wrote this manually, or your IDE formats it like this? I've always disliked this approach for importing multiple objects (unless it's a very large number) as it works against writing code quickly

app/api/services/core/patch_params.py
```diff
@@ -0,0 +1,12 @@
+import dataclasses
+from typing import Generic, TypeVar
+
+ResourceType = TypeVar("ResourceType")
```
chouinar: Is resource the right word? I didn't think of the request params on reading this at first glance, I'd say adding a `bind=` to it would help, but there's no dataclass class to bind to unfortunately.

app/api/services/core/patch_params.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+@dataclasses.dataclass
+class PatchParams(Generic[ResourceType]):
+    """Parameters for patching a resource."""
+
+    resource: ResourceType
+    fields_to_patch: list[str]
```
chouinar: I wonder if a utility generator method would be convenient here, something like:
```py
def get_fields_to_patch(self) -> Iterable[Tuple[str, Any]]:
     for field in self.fields_to_patch:
          value = getattr(self.resource, field)
          yield field, value
```
Might make the typing annoying, but I'm not sure how the existing approach doesn't hit typing issues.

app/api/route/schemas/user_schemas.py
lorenyu: oh i was just following [google's python import guidelines](https://google.github.io/styleguide/pyguide.html#22-imports), i think it often improves readability to see the class/function in the context of the module (e.g. `module.class`) especially if the class/function has a name that's a bit generic i.e. not entirely obvious what context it's for, and so i think something like `user_service.CreateUserParams` is clear that the CreateUserParams is part of the user_service module. otherwise it can potentially be confusing since we also have `user_models.User` in the `db.models` package.

i used to do something like `from api.services.user_service import CreateUserParams` but i've (very recently) started to appreciate the formality of google's guideline.

as an aside, [in this comment on jesse's PR](https://github.com/navapbc/template-application-flask/pull/56#discussion_r1060731741) i noticed that by importing the module rather than specific items in the module, it also helps avoid accidentally re-exporting things that weren't meant to be re-exported, which can avoid importing the wrong thing from the wrong module

app/api/route/schemas/user_schemas.py
lorenyu: that's true, but i think `Mapping` is a bit more general, since i don't really care whether `data` is an actual dict or just a dict-like object that supports the `**` unpacking operator

app/api/services/core/patch_params.py
lorenyu: good point now that i think of it resource isn't the right word. what about just leaving it generic as `T = TypeVar("T")`

yeah i agree it would be nice to add a `bind=`, but i didn't see a strong need to have a base class for the various params dataclasses.

app/api/services/core/patch_params.py
lorenyu: i like that idea! could i create a ticket to do that in a future PR?

app/api/services/users/models.py
lorenyu: good point. how about we put the objects next to the appropriate service? so, `CreateUserParams` lives in `create_user.py` and `PatchUserParams` lives in `patch_user.py`?

app/api/services/users/models.py
lorenyu: if you look in `create_user.py` below, we assert that certain fields aren't `None`, but don't do those assertions in `patch_user.py` since they can be `None` in the patch context. what do you think about that approach? to me it feels like the simplest one to understand and still has benefit of static analysis, since the asserts effectively rule out the None values in the context of create_user. I considered writing out separate classes but that will have a lot of duplicate code, and generating the separate class dynamically doesn't give us the benefit static analysis.

chouinar: It would only help in that particular function. Let's assume that right after you do all of those asserts you do:

```py
@dataclass
class MyModel:
     field: Optional[str] = None

assert my_model.field is not None
persist_my_model(my_model)

def persist_my_model(my_model: MyModel) -> None:
      my_field: str = my_model.field
```
mypy will actually complain because `field` is nullable inside the `persist_my_model` function despite you doing an assert because you could call this persist function from other places. You'd have to do these asserts in every separate part of the code, so they're actually doing nothing.

app/api/route/schemas/user_schemas.py
chouinar: Won't it always be a dict in this case though?

app/api/route/schemas/user_schemas.py
chouinar: I guess I don't "get" those guidelines. It looks more like a consistency reason rather than for a purpose especially when it adds a lot of verbosity with little gain. Maybe I'm just stuck in my ways, but it feels like more effort (since the IDE doesn't inherently do it) and more confusing to me (since every reference to a class/function requires a much longer signature). Not saying there aren't scenarios where I like the pattern, but it doesn't feel right

app/api/services/users/models.py
chouinar: That might be a bit clearer

app/api/services/users/models.py
chouinar: I wonder if there is a way to solve this with generics.


```py
K = TypeVar("K") # This can be either `Optional` or whatever the opposite of that is (Any?)

class MyModel[Generic[K]):
     param1: K[str] # So this is either Optional[str], or DummyWrapper[str] (which is the same as str)
     param2: Optional[str] # this is always optional even in post, so no need to do anything
     param3: Optional[str]
```

We might need to not use dataclasses in this case and instead use pydantic as it handles defaults cleaner.

lorenyu: good point about potentially needing assert more than once in a nested function call. that makes me think: given that the lifespace of the "params" objects (e.g. CreateUserParams) is meant to be pretty short-lived (it's just meant to translate from json in the api request to the parameters the service method needs in a type-safe manner, i wonder if a more pythonic way might be to actually put the parameters in the service method as parameters, since that's really what they are. in other words:

```
def create_user(first_name: str, last_name: str, ...):
    ...

def patch_user(first_name: Optional[str] = None, last_name: Optional[str] = None, ...):
    ...
```

then rather than passing in full params objects, we pass in the expanded version of the params objects as kwargs, which solves the problem you described around needing to assert multiple times.

lorenyu: yeah the generic idea is interesting, i had considered using python metaprogramming functionality as well, but worried that it's too complicated

app/api/services/users/models.py
lorenyu: cool i'll work on that today

app/api/route/schemas/user_schemas.py
lorenyu: yeah i believe so. i can change it back, dict is simpler and we don't need the generalization

app/api/route/schemas/user_schemas.py
lorenyu: What do you think about the last aside point? for example, if we did `from api.services.user_service import CreateUserParams` here in this `user_schemas.py` file, then because there's no true concept of public/private in python, what happens is that now the `user_schemas` module is actually re-exporting `CreateUserParams`, so someone can easily accidentally do `from user_schemas import CreateUserParams`, which is precisely what happened in Jesse's PR. I believe even the IDE autosuggestions could potentially suggest the incorrect import. And it's a pretty difficult error to spot in review since as a reviewer I usually just skim over the imports.

app/api/services/users/models.py
chouinar: If it's reasonable to make the type hint as simple as needing to mark a class as patch or not, then I don't think it'd be TOO complex to understand, but really depends on how much needs to be done for mypy to be happy.

I'm really surprised that there isn't a happy solution online, most of the time people just say to work with dictionaries and I really don't want to see those systems. Best solution I've seen is just defining the model twice which works, is clear, just mildly error-prone/tedious for large objects.

app/api/route/schemas/user_schemas.py
chouinar: I see that being a potential issue, although I've only seen it when refactoring (a function/class used to be in a file, got moved elsewhere, and imports daisy-chain without being noticed).

I wonder if there's a flake8 or other linter that can catch that? I found https://pypi.org/project/flake8-import-graph/ - but that seems to require explicit rules (eg. model files can't import sqlalchemy)

app/api/services/core/patch_params.py
lorenyu: Created this ticket https://github.com/navapbc/template-application-flask/issues/61

app/api/services/users/models.py
lorenyu: yeah i'm surprised too.

i also wonder if we're overcomplicating things. is it really that bad to just have two separate classes e.g.

```
class CreateUserParams:
    first_name: str
    last_name: str

class PatchUserParams:
    first_name: Optional[str]
    last_name: Optional[str]
```

or just define the params directly in the service methods

```
def create_user(
    first_name: str,
    last_name: str
):
    ...

def patch_user(
    first_name: Optional[str],
    last_name: Optional[str]
):
    ...
```

it's very simple and straightforward and to the point.

i often follow the rule of thumb of "three", where i only try to generalize and abstract once we duplicate the same thing 3 times. but create and patch is only two.

thoughts?

chouinar: I think two separate classes is fine. Whatever time we might spend on this to solve with a clever solution at this point probably offsets the small gains from anyone using it later.

The function approach is fine up to a point, if we're talking 3-4 params sure, but if we have a 20+ param then I'd probably make an object to hold all the params for the function anyways (or you can use kwargs which is literally back to being the dict approach).

I think at the end of the day, what I'd do if I were building an API with full control is just never let any of these models get too big. Probably avoid anything like the Mass PATCH application endpoint and instead make those a bit more targeted if possible (per-page on the front-end?)

lorenyu: ok pushed up a change with two separate params classes, and also moved them to the create_user and patch_user files respectively so the params live with the service methods.

note that `patch_user.py` imports `create_user.RoleParams`, does that feel ok or weird to you?

app/api/services/core/patch_params.py
lorenyu: As part of the last commit I went ahead and did this in a function called `fields_to_patch`, it's not an instance method though, i ended up trying one of @zelgadis 's earlier ideas of using a "unset"/"missing" sentinal value. (looks like under the hood Marshmallow does something similar, and I wouldn't be surprised if Pydantic also has something similar). thoughts on this approach? i like how explicit it is because then it makes it possible and clear to do things like

```
class FooParams:
    non_null: str | Missing
    can_be_null: str | None | Missing
```

app/api/services/users/models.py
lorenyu: @chouinar doing some more reading, i think i found something promising.

what do you think about [using a TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) instead of a dataclass?

we could do

```
user_fields = {
    "first_name": str,
    "middle_name": str,
    "last_name": str,
    "phone_number": str,
    "date_of_birth": date,
    "is_active": bool,
    "roles": list[RoleParams],
}

CreateUserParams = TypedDict('CreateUserParams', user_fields)
PatchUserParams = TypedDict('PatchUserParams', user_fields, total=False)
```

the `total=False` marks all the fields as `NotRequired`

chouinar: Let me test this a little, seems promising.

chouinar: A few notes from tinkering with it locally.

My IDE is able to auto-complete when I type `user["` it'll give me a list of options. It also shows type hints properly for `z = user["x"]`.

A bit of tinkering with mypy, it behaves as I would expect. Extra fields in the dictionary give an error. Typing still works (total=False allows it to not appear, but if it is there it has to follow typing).

I'm not finding any issues, I'm onboard with this idea.

lorenyu: i tinkered with it a bit too and ran into one issue, i'm wondering if you ran into it too.

when i use the functional syntax for creating the TypedDict

```
user_fields = {
    "first_name": str,
}

CreateUserParams = TypedDict('CreateUserParams', user_fields)

def create_user(params: CreateUserParams):
    params["first_name"] # this raises a type error
```

but when I use the class syntax like below it's ok

```
CreateUserParams(TypedDict):
    first_name: str

def create_user(params: CreateUserParams):
    params["first_name"] # this doesn't raise a mypy error anymore
```

i was hoping the functional syntax could work because then we could reuse the same dictionary to define `CreateUserParams` and `PatchUserParams` except `PatchUserParams` would have `total=False`

did you run into that at all?

otherwise we could just define two types separately with a bit of duplication

chouinar: Doing it the functional way, I'm not getting any different behavior. What type error are you getting?

lorenyu: i'm getting

> TypedDict "TypedDict" has no key "first_name"

image

lorenyu: It looks like it needs to use a literal dict, I can't assign a dict to a variable and then pass in the variable.

e.g. this works
```
Point2D = TypedDict('Point2D', {'x': int, 'y': int, 'label': str})
```

but this doesn't

```
fields = {'x': int, 'y': int, 'label': str}
Point2D = TypedDict('Point2D', fields)
```

lorenyu: Ok tried out TypedDict in a separate branch so we can compare the two: https://github.com/navapbc/template-application-flask/pull/62

Take a look and lemme know what you think.

lorenyu: Ok I updated the title and description of this PR now that we're using TypedDict instead of objects

lorenyu:
ready for a re-review. also, i put up a PR to document this in an ADR: https://github.com/navapbc/template-application-flask/pull/64

chouinar (approved):
LGTM!

PR #62, opened 2023-01-04
Switch to using TypedDict

lorenyu:
## Changes
* Switch to using TypedDict

## Context for reviewers
This PR builds off of https://github.com/navapbc/template-application-flask/pull/57. In [a PR thread on that PR](https://github.com/navapbc/template-application-flask/pull/57#discussion_r1061822399) @chouinar and I were discussing the possibility of getting some type safety by using TypedDict instead of with dataclasses. The pro is that the implementation is a lot simpler, since we don't need to define a `missing`/`omitted` sentinal value for class values, as the absence of a dictionary key is sufficient after setting `total=False` in the type definition. The con is that we don't get the runtime type safety of dataclass constructors, so if for some reason what gets loaded from `UserSchema` doesn't match the type definition of `CreateUserParams`, I don't think it will catch it until later.


chouinar:
@lorenyu - What do you mean by `The con is that we don't get the runtime type safety of dataclass constructors` - dataclasses don't do that (Pydantic does).

```py
@dataclass
class Something:
    x: str
    y: Optional[str]

def my_func():
    my_thing = Something(4, 5)
    print(my_thing)
```

will run perfectly fine, mypy will complain, but at runtime it has zero issue with it.

app/api/services/users/patch_user.py
```diff
@@ -1,25 +1,22 @@
...
-    is_active: bool | Missing = missing
-    roles: list[RoleParams] | Missing = missing
+class PatchUserParams(TypedDict, total=False):
+    first_name: str
+    middle_name: str
+    last_name: str
+    phone_number: str
+    date_of_birth: date
+    is_active: bool
+    roles: list[RoleParams]
```
chouinar, line 19: (stream of consciousness thoughts ending with - "this is fine")

Did we not want to do the define-once approach with them differing on `total=False`? Ah, wait, the PEP explicitly states it has to be done that way as it's only there for backwards compatibility.

I did try tinkering with inheritance slightly. You can't do a base class without any inheritance and then try to do `class Subclass(BaseClass, TypedDict)`, multiple inheritance is explicitly blocked for TypedDict.

You also can't do:
```py
class BaseExample(TypedDict):
    x: str
    y: Optional[str]

class ExampleA(BaseExample, total=False):
    pass

base: BaseExample = {"x": "asdasdasd"}
something: ExampleA = {"x": "asdasd"}
```
as the `total=False` doesn't get set - and I don't know enough about the class-inheritance bits in the TypedDict typing meta functions it calls to know if there is a clever workaround (probably not, and probably bad to consider).

So, yeah, it'd need to be duplicated, but I think that's just how it needs to be regardless of the technical approach (unless you want to do a lot of not-null checks for POST).

chouinar (approved):
I like this approach better than dataclasses or any of the others we tried. I was mostly looking for:
* Strong typing
* Ease of working with (IDE being able to give auto-completes, and info about variables in the object)
* Easy to understand
* Not duplicating post/patch models

That last one I think we've realized is just not plausible with any approach we've looked at, so, 3/4 is good enough. TypedDict handles the others simply enough, and also avoids needing the boilerplate of things like the post_load functions (since we're not actually restructuring it at all).

lorenyu:
> @lorenyu - What do you mean by `The con is that we don't get the runtime type safety of dataclass constructors` - dataclasses don't do that (Pydantic does).
> 
> ```python
> @dataclass
> class Something:
>     x: str
>     y: Optional[str]
> 
> def my_func():
>     my_thing = Something(4, 5)
>     print(my_thing)
> ```
> 
> will run perfectly fine, mypy will complain, but at runtime it has zero issue with it.

I guess dataclasses isn't perfect either, but as an example, with typeddict the following code will type-check fine but will result in a KeyError 'typo' in print_foo since at runtime `foo` only has `bar` as a key, not `typo`.

```
class FooSchema(Schema):
    bar = fields.String()

class Foo(TypedDict):
    typo: str

foo: Foo = FooSchema().load({"bar": "baz"})

def print_foo(foo: Foo) -> None:
    print(foo["typo"])

print_foo(foo)
```

but with dataclasses the following code will error out just slightly earlier at the Foo constructor in the @post_load hook with "Exception has occurred: TypeError Foo.__init__() got an unexpected keyword argument 'bar'"

```
class FooSchema(Schema):
    bar = fields.String()

    @post_load
    def make_foo(self, data, **kwargs) -> "Foo":
        return Foo(**data)

@dataclass
class Foo:
    typo: str

foo = FooSchema().load({"bar": "baz"})

def print_foo(foo: Foo) -> None:
    print(foo.typo)

print_foo(foo)
```

i don't think it's a big con but just wanted to call it out

app/api/services/users/patch_user.py
lorenyu: appreciate the stream of consciousness!

i like this approach best too, i'll go ahead and merge it into the other PR

Issue #37, opened 2022-12-13
Migrate poetry dev dependencies to new group syntax


lorenyu:
* https://python-poetry.org/blog/announcing-poetry-1.2.0/
* https://python-poetry.org/docs/master/managing-dependencies/

PR #65, opened 2023-01-06
Add make test-watch

### LINKED ISSUE #26: Add make test-watch
I noticed we have `make test` and `make test-coverage` but not `make test-watch`. Could be useful for developer productivity

lorenyu:
## Ticket

Resolves #26 

## Changes
see title

## Context for reviewers
I know you don't really use test-watch as part of your workflow but since it's only a dev dependency and doesn't add much additional complexity to the codebase I was thinking we could add it (I personally would like to use it).

## Testing
Ran `make build` then `make test-watch`

### DIFF
--- a/app/pyproject.toml
+++ b/app/pyproject.toml
@@ -35,6 +35,9 @@ Faker = "^14.2.0"
 factory-boy = "^3.2.1"
 bandit = "^1.7.4"
 
+[tool.poetry.group.dev.dependencies]
+pytest-watch = "^4.2.0"
+
 [build-system]
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
--- a/app/Makefile
+++ b/app/Makefile
@@ -124,6 +124,9 @@ db-migrate-heads: ## Show migrations marked as a head
 test:
 	$(PY_RUN_CMD) pytest $(args)
 
+test-watch:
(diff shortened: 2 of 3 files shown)


app/pyproject.toml
```diff
@@ -35,6 +35,9 @@ Faker = "^14.2.0"
 factory-boy = "^3.2.1"
 bandit = "^1.7.4"
 
+[tool.poetry.group.dev.dependencies]
```
lorenyu, line 38: this is the new way of doing dev dependencies. i didn't bother migrating everything over in this PR since we have a separate ticket for that #37

chouinar (approved):
LGTM! I did a quick test of it and it worked as I would expect. Heads up that there are a lot of deprecation warnings that I saw printed out regarding Marshmallow (unrelated to this) when running the tests:

```
.venv/lib/python3.11/site-packages/marshmallow/fields.py:1256
  /Users/michaelchouinard/workspace/template-application-flask/app/.venv/lib/python3.11/site-packages/marshmallow/fields.py:1256: RemovedInMarshmallow4Warning: Passing field metadata as keyword arguments is deprecated. Use the explicit `metadata=...` argument instead. Additional metadata: {'description': 'The users date of birth'}
    super().__init__(**kwargs)
```

lorenyu:
Thanks, I'll look into the deprecation warnings

PR #68, opened 2023-01-06
Combine and parametrize user route tests

### LINKED ISSUE #45: Simplify test assertions in test_user_route
I wonder if validate_match functions are necessary or if we can rely on built in python operators

lorenyu:
## Ticket

Resolves #45 

## Changes
* Use public interfaces for testing
* Replace custom validation function with built in pytest equality assertions
* Combine tests and parametrize them

## Context for reviewers
I took a pass at combining some of the tests and parametrizing them instead. This should reduce the amount of executable test code in the tests so that developers can focus on the different data parameters.

I also wanted to rely more on the built in pytest assertions which have very developer friendly output, so I attempted to rewrite the assertions to use built-in equality operators rather than custom validation functions. This should hopefully reduce the amount of code a reader needs to look at, and also should allow a reader looking at the test to be able to fully and more precisely understand what the API endpoints do.

Finally, I was looking to use only public interfaces for testing, so to create the users rather than using the user factory I opted to use just use the POST /users endpoint to create users and then to use the GET /users endpoint to validate that things were persisted.

## Testing
`make test`


app/api/route/schemas/user_schemas.py
```diff
@@ -14,10 +14,6 @@ class RoleSchema(request_schema.OrderedSchema):
         user_models.RoleType, description="The name of the role", by_value=True
     )
 
-    # Output only fields
```
lorenyu, line 19: decided to remove the created_at/updated_at attributes from the role in the openapi schema, since they aren't really a primary resource

app/api/db/models/user_models.py
```diff
@@ -28,7 +28,9 @@ class User(Base, IdMixin, TimestampMixin):
     date_of_birth: date = Column(Date, nullable=False)
     is_active: bool = Column(Boolean, nullable=False)
 
-    roles: Optional[list["Role"]] = relationship("Role", back_populates="user")
+    roles: Optional[list["Role"]] = relationship(
+        "Role", back_populates="user", order_by="Role.type"
```
lorenyu, line 32: the `order_by` makes the order of the roles deterministic, something I caught when adding the various combinations of patching roles. previously if a user started out with `USER` role and added `ADMIN` the order will be `USER`, `ADMIN`, this ensures the order of the roles is always the same

app/tests/api/route/test_user_route.py
```diff
@@ -52,216 +67,167 @@ def validate_param_match(key, request, response, db_record):
...
+    assert get_response.status_code == 200
+
+    get_response_data = get_response.get_json()["data"]
+    assert get_response_data == expected_response
+
+
+@pytest.mark.parametrize(
+    "request_data,expected_response_data",
+    [
+        pytest.param(
```
chouinar: Thoughts on instead defining this with the scenarios separated to keep it from being so nested - similar to the idea of not nesting if-statements so deep, not nesting data (at least in appearance) so deep?

```py
SCENARIO_A = pytest.param(...)
SCENARIO_B = ...
SCENARIO_C = ...

@pytest.mark.parametrize(
    "request_data,expected_response_data",
    [ SCENARIO_A, SCENARIO_B, SCENARIO_C]
)
```

At least for any of these very large scenario definitions

app/tests/api/route/test_user_route.py
```diff
@@ -18,6 +20,19 @@
 }
 
 
+def powerset(iterable):
+    "powerset([1,2,3]) --> [] [1] [2] [3] [1,2] [1,3] [2,3] [1,2,3]"
```
chouinar: I think this was supposed to be a comment, not just a floating string.

Could you add a bit more detail to this? I think I get the idea that it's all the combinations, but more clarity always helps, especially an example like "if you run a test using this like X, it'll run combinations A, B, C". Effectively explaining some of the uses of this function you did below.

I know utility files are a bit of a questionable implementation approach (I like them personally in certain cases), but this does feel like something that would be good to put in some sort of `test_utils.py` for all the tests to potentially use.

app/tests/api/route/test_user_route.py
```diff
@@ -52,216 +67,167 @@ def validate_param_match(key, request, response, db_record):
...
+                "last_name": ["Not a valid string."],
+                "phone_number": ["Not a valid string."],
+                "date_of_birth": ["Not a valid date."],
+                "is_active": ["Not a valid boolean."],
+                "roles": ["Not a valid list."],
+            },
+            id="invalid types",
+        ),
+        pytest.param(
+            base_request | {"roles": [{"type": "Mime"}, {"type": "Clown"}]},
```
chouinar: Not an issue here, just curious on how you'd approach something. One I've seen especially with this union operator is accidentally overriding the `base_request`.

If you instead do `base_request |= { ... }` then it updates base_request but since that's a module level value, it changes it for all tests causing the tests to interfere with each other. Now, that's not an issue here, but if a test were written that did that then it might be difficult for a developer to know "why" it's broken as test A is causing test B to fail, and only when you run them all, not individually.

I worked around this issue by changing:
```py
base_request = { ... }
```
into
```py
def get_base_request():
     return { ... }
```

That way it's generated fresh each test. Thoughts?

chouinar (commented):
Added a few questions to discuss, but looks good

app/tests/api/route/test_user_route.py
lorenyu: yeah what i've done in the past is something in between, rather than a separate top level variable for every scenario, i'd do something like

```
test_data = [
    (..., ...),
    (..., ...),
    (..., ...),
]
@pytest.mark.parametrize("foo,expected", test_data)
```

i pushed up a change that does this

app/tests/api/route/test_user_route.py
lorenyu: i moved it to `tests/api/util/parametrize_utils.py` and added some docstrings, take a look and lemme know what you think

app/tests/api/route/test_user_route.py
lorenyu: yeah i had the same thought. i made a `base_request` fixture, and also made a `get_base_request()` function for situations where we need to call it during collection (e.g. to use in pytest.mark.parametrize). i think this should solve the issues you're talking about

chouinar (approved):
LGTM!

PR #71, opened 2023-01-10
Fix ability to access server when running in container

### LINKED ISSUE #70: App isn't accessible when running in docker container

lorenyu:
## Ticket

Resolves #70

## Changes
* Add container name when running from docker-compose
* Add make login to log into docker container
* Add AppConfig.host to set host that Flask app listens to, defaulting to 127.0.0.1, the secure default for Flask
* Override HOST in Dockerfile to 0.0.0.0 to allow external connections when running in a container

## Context for reviewers
Since PR #51 was merged, the app was broken when running in docker. This was because connexion when running the app in __main__.py, connexion defaults host to 0.0.0.0, but flask defaults host to 127.0.0.1, so the app within the container wasn't allowing incoming requests external to the container.

## Testing
* Tested with `make start` which runs docker-compose up, saw from logs that host is 0.0.0.0 and hit localhost:8080/health in a browser
* Tested with `poetry run python -m api` which runs locally, saw from logs that host is 127.0.0.1 and hit localhost:8080/health in a browser to see that it works

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -26,6 +26,10 @@ RUN poetry install --no-root
 # Copy application files.
 COPY . /srv
 
+# Set the host to 0.0.0.0 to make the server available external
+# to the Docker container that it's running in.
+ENV HOST=0.0.0.0
+
 # Install application dependencies.
 # https://python-poetry.org/docs/basic-usage/#installing-dependencies
 RUN poetry install
--- a/app/Makefile
+++ b/app/Makefile
@@ -175,6 +175,9 @@ create-user-csv:
 # Miscellaneous Utilities
 ##################################################
 
+login: start ## Start shell in running container
+	docker exec -it $(APP_NAME) bash
+
 # Pauses for 5 seconds
 sleep-5:
 	sleep 5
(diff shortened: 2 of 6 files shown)


jamesbursa (approved):
Looks good, but there's a linting error from bandit:

> Issue: [B104:hardcoded_bind_all_interfaces] Possible binding to all interfaces.
   Severity: Medium   Confidence: Medium
   CWE: CWE-605 (https://cwe.mitre.org/data/definitions/605.html)
   Location: ./api/app_config.py:10:16
   More Info: https://bandit.readthedocs.io/en/1.7.4/plugins/b104_hardcoded_bind_all_interfaces.html
9	    # See https://flask.palletsprojects.com/en/2.2.x/api/#flask.Flask.run
10	    host: str = "0.0.0.0"
11	    port: int = 8080

lorenyu:
@jamesbursa oh hmm any suggestions on how to deal with that? is there a better way to allow external connections that doesn't involve 0.0.0.0?

lorenyu:
@jamesbursa I switched to defaulting to 127.0.0.1, and set the HOST environment variable to 0.0.0.0 in the Dockerfile, which seemed like the most precise way to say "use 0.0.0.0 when running in a container"

jamesbursa (approved):

PR #74, opened 2023-01-12
Refactor db session management to eliminate dependency on global scoped_session

### LINKED ISSUE #44: Remove session_scope context manager now that SQLAlchemy 1.4 supports it by default
🔒  https://nava.slack.com/archives/C03G1SWD9H7/p1671137100224249

lorenyu:
## Ticket

Resolves #44 

## Changes
* Refactor db session management to eliminate dependency on global scoped_session

db module changes
* Add DB class
* db.init creates an instance of DB class instead of setting global scoped_session
* db.init_app sets db instance on flask_app.extensions dictionary
* get_db gets db instance from flask_app.extensions dictionary

db testing changes
* Rename test_db to db since test context is already assumed
* Move creation of test db schema to db_testing module function create_isolated_db
* Add isolated_db to create isolated schema for individual test
* Route/flask_app tests no longer assume db changes rolled back per test,
   instead tests that require isolation should use isolated_db fixture
* Rename test_db_non_session to empty_schema to create isolated empty schema used for test_migrations
   and moved to conftest.py to be with other similar fixtures
* Refactor factories.py to use None check of module level variable instead
   of DB_FACTORIES_DISABLE_DB_ACCESS environment variable

other related changes
* Simplify factories 
* Remove app.db_session, db_session_raw, and db_session context manager functions
* Do not set db session on flask.g request context
* Remove option to set session factory in create_app
* Add documentation about db client and db testing

Unrelated changes
* Clear terminal on each test run when using pytest-watch
* Fix pyproject.toml parse error for pytest-watch

## Context for reviewers
This PR is a big one that aims to simplify and db session management by removing the global singleton session_scope object and instead saving the db engine directly on the flask app extenstions dictionary. Furthermore, the flask global request context `g` object is no longer used, opting instead for request handlers to explicitly instantiate a new session per request rather than relying on the request per thread behavior of session_scope. In pulling on this thread, db testing fixtures and test factories also needed to be updated.

This took me a while to untangle since there were so many interrelated runtime possibilities (create_app accepting a db_session_factory, the factories module having it's own session_scope object that referenced the db module's session_scope object), so I think it's important that we remove a lot of those parameters to reduce branching complexity. For example, I removed the ability to pass in a db_session_factory to create_app, opting instead to force tests to explicitly require an isolated_db fixture if they truly need to have their db state isolated from other tests. As such, test_user_route does end up creating users in the database. The behavior of discarding the postgres schema at the end still remains, so the test suite session is still isolated.

**note:** I probably broke the create-user-csv = "api.scripts.create_user_csv:main" script, but I'm thinking fixing it could be in a future PR. I'm thinking we should leverage built in flask cli commands rather than creating our own, since I notice we have duplicate code in multiple __main__.py files, and moreover that code isn't tested by the unit test suite.

cc'ing @jamesbursa @kyeah @doshitan for visibility since this PR is related to previous work they've done

## Testing
ran `make check`
ran `make start` and checked the /health page


app/api/db/__init__.py
```diff
@@ -1,60 +1,159 @@
...
-    verify_ssl(conn_info)
-
-    # Explicitly commit sessions — usually with session_scope. Also disable expiry on commit,
-    # as we don't need to be strict on consistency within our routes. Once we've retrieved data
-    # from the database, we shouldn't make any extra requests to the db when grabbing existing
-    # attributes.
-    session_factory = scoped_session(
-        sessionmaker(autocommit=False, expire_on_commit=False, bind=engine)
-    )
+class DB:
```
chouinar: I feel like the name `DB` is too raw. Working with that, I'd assume I'm using some very low level DB operator and not the SQLAlchemy session manager that it really is. What about something like DBManager or DBClient?

app/api/db/__init__.py
```diff
@@ -1,60 +1,159 @@
...
+            # TODO add check_migrations_current to config
+            # if check_migrations_current:
+            #     have_all_migrations_run(engine)
+
+
+def init() -> DB:
+    return DB()
+
+
+def get_db(app: Flask) -> DB:
```
chouinar: Should this function live in this file? If I were working on a backend script that also lives in this repo, it would be odd to pull in a function expecting a Flask app.

app/api/db/__init__.py
```diff
@@ -1,60 +1,159 @@
...
-from sqlalchemy.engine import Engine
-from sqlalchemy.orm import scoped_session, sessionmaker
+from flask import Flask
+from sqlalchemy.orm import session
 
 import api.logging
 from api.db.db_config import DbConfig, get_db_config
-from api.db.migrations.run import have_all_migrations_run
+
+FLASK_EXTENSION_KEY = "db"
```
chouinar: Maybe `FLASK_DB_EXTENSION_KEY` to distinguish it from any other one?

app/api/route/user_route.py
```diff
@@ -27,8 +28,9 @@ def user_post(user_params: users.CreateUserParams) -> dict:
     """
     logger.info("POST /v1/user")
 
-    with api_context_manager() as api_context:
-        user = user_service.create_user(user_params, api_context)
+    db = api.db.get_db(current_app)
+    with db.get_session() as db_session:
```
chouinar: Do you think it would be useful to have a shortcut method that lets you just do `with api.db_session()...` so you don't need to unpack as much?

app/api/scripts/util/script_util.py
```diff
@@ -1,8 +1,9 @@
+# TODO use built in @flask.cli commands so that we can reuse flask app and no longer need this file
```
chouinar, line 1: I'm not familiar with Flask CLI, but this is explicitly for backend scripts separate from the API itself, like a lot of our backend processes from PFML (payments, file imports, and so on).

app/api/db/__init__.py
```diff
@@ -1,60 +1,159 @@
...
+                },
+            )
+            verify_ssl(conn_info)
+
+            # TODO add check_migrations_current to config
+            # if check_migrations_current:
+            #     have_all_migrations_run(engine)
+
+
+def init() -> DB:
```
chouinar: Thoughts on changing this to the below to avoid someone forgetting to check the connection where it makes sense to do so? I think outside tests we'd generally want to do that, right?

```py
def init(verify_connection: bool = True) -> DB:
    db = DB()
    if verify_connection:
         db.check_db_connection()

     return db
```

chouinar (commented):
Just some structure thoughts and ideas

app/api/db/__init__.py
lorenyu: I like both of those names, do you prefer one over the other? if not i'll go with DBClient since it's slightly shorter and also consistent with other external services (and i think db can be thought of as a special case of another service, in this case RDS)

app/api/db/__init__.py
lorenyu: my understanding is that while flask is mostly for the web use case, that it also supports custom CLI commands (https://flask.palletsprojects.com/en/2.2.x/cli/#custom-commands) that appears to be the same as what we're doing manually with the individual script files in the scripts directory, so it'd be nice to consolidate and use the flask app as the central "app" that we run in both the "web app" and "batch job" contexts. thoughts?

app/api/db/__init__.py
lorenyu: i was thinking that this is a constant private to this file e.g. it's the FLASK_EXTENSION_KEY for the api.db module. what if instead i prefixed the constant with _ like `_FLASK_EXTENSION_KEY`, would that make it clearer?

app/api/route/user_route.py
lorenyu: It would be more convenient for the developer / save one line of code, but one feature I like about the current approach is that it makes it explicit that this file depends on the `api.db` module. If we switched it to `api.db_session` then it's not clear from looking at the imports that this module depends on functionality in `api.db`. I was thinking that longer term it could be beneficial to add a linter/test/check that helps enforce architectural patterns, such as modules in `services` should not import modules in `routes`, etc. so having dependencies by explicit is a nice bonus IMO. What do you think?

app/api/db/__init__.py
lorenyu: Sounds good to me. I was on the fence about it, since I was generally hoping to avoid parameters in the init function. The previous init function with 3 parameters made it harder for me to reason about all the various entry points, whereas putting the the extra calls in the places that needed them made it easier for me to understand. That said I think you're right in that it actually simplifies the number of lines in `__main__` and only adds complexity to the test suite, which I'm more ok with.

app/api/scripts/util/script_util.py
lorenyu: Kind of related to the [other thread](https://github.com/navapbc/template-application-flask/pull/74#discussion_r1068634113) so maybe we can just continue on that thread, but I was surmising that Flask custom CLI commands could replace our custom/homegrown entrypoints for backend processes.

app/api/db/__init__.py
lorenyu: Another advantage with using the Flask custom CLI command framework rather than making our homegrown entry point is that we'd be able to leverage blueprints. In the future I'd like us to build reusable blueprints for things like "doc upload endpoints" or "user account management" (change address, change name, etc), and those blueprints could come with their own CLI commands like "disable user account".

app/api/route/user_route.py
chouinar: What about importing as `import api.db as db`, I actually at first assumed that db was some object in the api module, when the api isn't actually relevant here.

Could also shorten this as 
```
with db.get_db(current_app).get_session() as db_session:
```
Which still feels very verbose.

app/api/db/__init__.py
chouinar: Wasn't aware that Flask supported CLI commands like that, seems a little odd, but I do like click over something like argparse.

Just brainstorming, what flask features would we be able to leverage with running via CLI? I know flask manages several contexts, but any other relevant utils or features that we might not normally use for the API?

I'd be curious to see some scripts built out using it as those examples are very basic. Something similar to the ECS tasks on Mass (at least the entry-point). I do think having some utilities/patterns for how to build a task, and instrument them would be useful. Porting over something like the `background_task` from the Mass project which sets up logging, metrics, task name, and so on.

lorenyu: brainstorming some features that we can use:

* application_context, like you suggested, which can be used to manage extensions that we install in the future, and custom extension like the DB client we added in this PR. so CLI scripts can do `db = api.db.get_db(current_app)`
* [debug mode](https://flask.palletsprojects.com/en/2.2.x/cli/#debug-mode) interactive debugger and reloader
* [registering CLI commands with blueprints](https://flask.palletsprojects.com/en/2.2.x/cli/#registering-commands-with-blueprints), like i described in the previous comment
* looks like some [flask extensions/plugins could add their own commands](https://flask.palletsprojects.com/en/2.2.x/cli/#plugins)

it also looks like we could still [build custom scripts using click](https://flask.palletsprojects.com/en/2.2.x/cli/#custom-scripts) and pass it the create_app factory if we wanted to go that route

> I'd be curious to see some scripts built out using it as those examples are very basic. Something similar to the ECS tasks on Mass (at least the entry-point). I do think having some utilities/patterns for how to build a task, and instrument them would be useful. Porting over something like the background_task from the Mass project which sets up logging, metrics, task name, and so on.

are you thinking we should do that in this PR (knowing that i might have broken the create_user_csv script) or would you be open to doing that in a future PR since this PR is not small

app/api/route/user_route.py
lorenyu: > I actually at first assumed that db was some object in the api module, when the api isn't actually relevant here.

that's a good point, we don't often import nested modules like this, so i can see the confusion.

i changed it to `import api.db as db` like you suggested, but then i also adopted the convention that instances of `DBClient` should be named `db_client` rather than `db`, to avoid confusion between the `db` client and the `db` module.

i also shortened the line as you suggested

lemme know your thoughts

lorenyu:
btw I went ahead and created some documentation in docs/app/database/ since I thought it might be good to have the docs and the code in the same PR

app/api/services/users/patch_user.py
```diff
@@ -23,69 +25,50 @@ class PatchUserParams(TypedDict, total=False):
...
+            # https://github.com/navapbc/template-application-flask/pull/51#discussion_r1053754975
+            raise apiflask.HTTPError(404, message=f"Could not find user with ID {user_id}")
+
+        for key, value in patch_user_params.items():
+            if key == "roles":
+                _handle_role_patch(db_session, user, patch_user_params["roles"])
+                continue
+
+            setattr(user, key, value)
+    return user
```
chouinar, line 49: Might be better to handle this with the persistence layer work that'll come later, but I'm curious if the services should be structured like this, since every single one will follow this pattern:

```py
with db_session.begin():
      # the service logic

return response
```

Feels like the session should begin a layer up, which also helps this be slightly less nested. When the persistence layer is added, I'd assume we don't want any DB logic here

Fine to wait until that is tackled to tinker with this more.

chouinar (approved):
LGTM! One structure question, but probably makes sense to handle it with the persistence work coming later.

app/api/services/users/patch_user.py
lorenyu: Yeah kind of agree, we can tinker with that in a subsequent PR

PR #84, opened 2023-01-18
Separate generic db adapter code from project specific db code

### LINKED ISSUE #85: Separate generic db adapter code from project specific db code
it'd be valuable to separate generic db adapter code (creating connections, sessions) from project specific db code (models, migrations) for maintainability and to make it easier for projects to pull in future updates to the template.

lorenyu:
## Ticket

Resolves #85

## Changes
* Pull out generic db adapter code into adapters/db package,
   leaving project-specific code (models/migrations) in api/db
* Pull out db/__init__.py code into two submodules client.py and flask_db.py

## Context for reviewers
There's no original code in this PR, just moving things around and adding additional comments/docstrings:

- I separated out the project-specific files like db/models and db/migrations from the core db adapter code that is agnostic to the project by moving the db adapter code to the api/adapters folder. (see https://medium.com/idealo-tech-blog/hexagonal-ports-adapters-architecture-e3617bcf00a0). By pulling out the generic code it should make things more maintainable and also make it easier for projects to pull in updates/improvements/bug-fixes to the generic package. Longer term we could also potentially pull this out into it's own pip module.
- Also, most of the db code was in one big \_\_init\_\_.py file, so I decided to just organize the code a bit to make the code easier to browse, as there are usually so many \_\_init\_\_ files in a project. I moved the DBClient class and connection related functions to db/client.py and moved the flask_extension related functions to db/flask_db.py. By separating out the framework specific code into flask_db this should make things more portable – if we ever want to support fastapi we need only add a fastapi_db.py submodule plugin.

## Testing
I ran `make check`
I also ran `make db-downgrade` and `make db-upgrade`


app/api/adapters/db/__init__.py
```diff
@@ -0,0 +1,31 @@
+"""
```
chouinar, line 1: Are there any other adapter files you plan to add besides for the db? Would `app/api/db/adapters/...` make more sense if not?

lorenyu: i plan to add a "storage" adapter for accessing local file storage and remote file storage (e.g. "aws s3" or "azure blob storage")

another common adapter is "queue" for accessing job queues, like rabbitmq or aws sqs, but i'm probably not going to prioritize adding that to the template anytime soon

chouinar: Got it, I assume you'll probably restructure the file utils we have into that which sort've work like that. Alright, no issue then

chouinar (approved):
LGTM!

PR #86, opened 2023-01-18
Add with_db_session decorator to flask_db module

### LINKED ISSUE #80: Add with_db_session and with_db_session_transaction decorators
Could be nice convenience to add decorators that automatically create a new session in a context manager (and optionally start transaction)

# Design options

could be two decorators `with_db_session` and `with_db_session_transaction` or could be one decorator with a parameter
`with_db_session(start_transaction: bool)`

lorenyu:
## Ticket

Resolves #80 

## Changes
see title

## Context for reviewers
Idea for using decorators as a convenience method to create db sessions in web requests.

### DIFF
--- a/app/api/adapters/db/flask_db.py
+++ b/app/api/adapters/db/flask_db.py
@@ -19,8 +19,12 @@ Usage:
     def health():
         db_client = flask_db.get_db(current_app)
 """
-from flask import Flask
+from functools import wraps
+from typing import Any, Callable, Concatenate, ParamSpec, TypeVar
 
+from flask import Flask, current_app
+
+import api.adapters.db as db
 from api.adapters.db.client import DBClient
 
 _FLASK_EXTENSION_KEY = "db"
@@ -51,3 +55,33 @@ def get_db(app: Flask) -> DBClient:
             db_client = flask_db.get_db(current_app)
     """
     return app.extensions[_FLASK_EXTENSION_KEY]
+
+
+P = ParamSpec("P")
+T = TypeVar("T")
+
+
+def with_db_session(f: Callable[Concatenate[db.Session, P], T]) -> Callable[P, T]:
+    """Decorator for functions that need a database session.
+
+    This decorator will create a new session object and pass it to the function
+    as the first positional argument. A transaction is not started automatically.
+    If you want to start to a transaction automatically, you can use
+    with_db_session_transaction.
+
+    Usage:
+        @with_db_session
+        def foo(db_session: db.Session):
+            ...
+
+        @with_db_session
+        def bar(db_session: db.Session, x, y):
+            ...
+    """
+
+    @wraps(f)
+    def wrapper(*args: Any, **kwargs: Any) -> T:
+        with get_db(current_app).get_session() as session:
+            return f(session, *args, **kwargs)
+
+    return wrapper
--- a/app/api/adapters/db/flask_db.py
+++ b/app/api/adapters/db/flask_db.py
@@ -13,7 +13,7 @@ Example:
     flask_db.init_app(db_client, app)
 
 Then, in a request handler, use the with_db_session decorator to get a
-database session for the duration of the request.
+new database session that lasts for the duration of the request.
 
 Example:
     import api.adapters.db as db
--- a/app/api/adapters/db/flask_db.py
+++ b/app/api/adapters/db/flask_db.py
@@ -1,23 +1,42 @@
 """
-This module functionality to extend Flask with a database client.
+This module has functionality to extend Flask with a database client.
 
-Usage:
-    # initialization in Flask's create_app()
+To initialize this flask extension, call init_app() with an instance
+of a Flask app and an instance of a DBClient.
 
+Example:
     import api.adapters.db as db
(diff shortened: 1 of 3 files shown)


app/api/route/user_route.py
```diff
@@ -22,21 +23,21 @@
 @user_blueprint.input(user_schemas.UserSchema)
 @user_blueprint.output(user_schemas.UserSchema, status_code=201)
 @user_blueprint.auth_required(api_key_auth)
-def user_post(user_params: users.CreateUserParams) -> dict:
+@flask_db.with_db_session_transaction
+def user_post(db_session: db.Session, user_params: users.CreateUserParams) -> dict:
```
chouinar: I definitely like this approach for the DB session with the APIs. Since we already have half a dozen decorators, it's clear that a lot is being injected into this method already. Since this method is already an entry-point, I think having the session there makes a lot of sense.

app/api/adapters/db/flask_db.py
```diff
@@ -51,3 +55,61 @@ def health():
...
+T = TypeVar("T")
+
+
+# https://github.com/python/mypy/issues/12280
+def with_db_session(f: Callable[[*Ts, db.Session], T]) -> Callable[[*Ts], T]:  # type: ignore
+    """Decorator for functions that need a database session.
+
+    This decorator will create a new session object and pass it to the function
+    as the first positional argument. The function's last parameter needs be named
+    "db_session". A transaction is not started automatically. If you want to start
```
chouinar: Should this say `The function's first parameter needs to be named`?

chouinar (commented):
I like this approach, I think it makes sense that the DB session would come in the entry-point function for a route.

app/api/adapters/db/flask_db.py
lorenyu: oh yeah i need to clean this PR up for sure

btw i realize that mypy doesn't yet support variadic type variables, so one downside with the decorator is that the type checking won't be as strong

lorenyu: oh nvm ignore the second part of that comment, i can use typing's ParamSpec and Concatenate

app/api/route/user_route.py
```diff
@@ -22,21 +22,22 @@
...
+    user = user_service.create_user(db_session, user_params)
 
-        logger.info(
-            "Successfully inserted user",
-            extra=get_user_log_params(user),
-        )
+    logger.info(
+        "Successfully inserted user",
+        extra=get_user_log_params(user),
+    )
```
lorenyu, line 37: I realized we still want to start the within the service method so that the service method can control how it completes it's work of flushing to the db. in general, when transactions should start/stop is determined by business logic, so the service should be responsible for that, not the framework.

i noticed that in this case when we started the transaction in the decorator, the create_user method hasn't completed it's work by this line of code, so get_user_log_params still shows user_id as being None

lorenyu:
Ok cleaned it up and ready for official review now

app/api/route/user_route.py
chouinar: You can work around that by doing a flush, or generating the uuid and setting it on the object yourself, but I think being explicit with transactions makes sense. Technically all of the Mass PFML code outside a few explicit commits did it exactly like that, but it never seemed to cause any issues.

Somewhat related, would you find it useful to be able to "register" parameters during a request (like the user ID here) and have them automatically attached to any subsequent logs? I added a utility like that in NJ that expands on https://github.com/navapbc/template-application-flask/blob/main/app/api/logging/log_attributes.py

chouinar (approved):
LGTM! Comment is mostly a tangential thought

app/api/route/user_route.py
lorenyu: Yeah I'd love to standardize on how we do logging to optimize for observable and debuggable code. Registering parameters sounds good, could you share the approach you did? Easy to adapt / port over?

Also, looks like [subclasses of Logger can override makeRecord](https://docs.python.org/3/library/logging.html#logging.Logger.makeRecord). I wonder if that could be of any use.

chouinar: I added this function to log_attributes:

```py

def add_request_log_attribute(key: str, value: Any) -> None:
    """
    For he duration of the current flask request, add
    a parameter that will appear in the extra section
    of any log statement. Useful for attaching an ID
    of a record relevant to a request.
    """

    if flask.has_request_context():
        request_log_attributes: dict[str, Any] = flask.g.get("request_log_attributes", {})
        request_log_attributes[key] = value

        flask.g.request_log_attributes = request_log_attributes
```

And added the following to `get_request_attributes`:

```py
        request_log_attributes = flask.g.get("request_log_attributes")
        if request_log_attributes:
            request_attributes |= request_log_attributes
```

Basically just continuing the pattern. It does use the flask `g` context, but since that's tied to the request context, I think it makes sense to keep using it like this since it's log info only relevant to the given request.

lorenyu: Oh I see! Neat, I'll give it a think 👍 
in the meantime i created this placeholder ticket for that idea https://github.com/navapbc/template-application-flask/issues/87

PR #88, opened 2023-01-19
Use Docker Compose V2

### LINKED ISSUE #33: Switch to docker compose v2
https://www.docker.com/blog/announcing-compose-v2-general-availability/

michael-wojcik:
## Ticket

Resolves #33

## Changes
* Updates all references to `docker-compose` to `docker compose`.


docs/app/README.md
```diff
@@ -51,6 +51,8 @@ See the [Makefile](./app/Makefile) for a full list of commands you can run.
 
 ## Docker and Native Development
 
+***NB.*** This template uses [Docker Compose V2](https://www.docker.com/blog/announcing-compose-v2-general-availability/) (viz. `docker compose` instead of `docker-compose`).
```
lorenyu: what does NB mean?

btw files in `docs/app` shouldn't reference "template" since these files will be copied over to projects and are meant to still be relevant in the context of the project. 

fyi documentation that is template specific should go into template-only-docs.

but in this case you'd just say something like, "this application uses..."

btw i don't we should document this here. if we want to say that docker compose v2 is required, i'd put that in docs/app/getting-started.md in the same line that says Docker Desktop is required. another option is to not document this anywhere, since Docker Desktop already includes docker compose v2, so there really isn't much value in knowing this

lorenyu (approved):
Looks good! Just had one comment about the documentation.

Also nitpick but could you clean up the placeholder text in the PR description

PR #89, opened 2023-01-19
Add test coverage for adapters/db package

lorenyu:
## Ticket

N/A

## Changes
* Add tests for db.init, get_connection and get_session
* Parametrize make_connection_uri test to get test missing branches (e.g. when username and port is "")
* Use get_app_security_schemes function rather than hardcoding in create_app

## Context for reviewers
I realized I didn't add tests for the db module so went ahead and did that, and while doing that I checked our test coverage and decided to patch up a couple of other areas related to the db package.

## Testing
CI now shows 100% for db package, and this remains true even if I just run the db tests e.g. with `args='k "test_db or test_flask_db"'`

image


chouinar (approved):
LGTM!

PR #94, opened 2023-01-21
Log query parameters for every request

lorenyu:
## Changes
see title

## Context for reviewers
URLs should be safe to log, as they are already logged in access logs, etc, so I think it would be valuable to add query params. This could be useful for API routes that include things like pagination in the query params

## Testing
CI


chouinar (approved):
LGTM! Makes sense to me

PR #96, opened 2023-01-21
Add function for adding custom data to current request logs

### LINKED ISSUE #95: Add ability to adding custom data to all of current request logs
One enhancement that was discussed was the ability to register parameters on the logger in a request context (see this [PR thread](https://github.com/navapbc/template-application-flask/pull/86#discussion_r1081410116))

lorenyu:
## Ticket

Resolves #95

## Changes
see title

## Context for reviewers
In [this comment thread](https://github.com/navapbc/template-application-flask/pull/86#discussion_r1081410116), Michael suggested adding the ability to add extra data that will be automatically added to all the logs for the current flask request. This PR adds that.

## Testing
CI


app/api/logging/flask_logger.py
```diff
@@ -47,13 +49,28 @@ def init_app(app_logger: logging.Logger, app: flask.Flask) -> None:
...
+    )
+
     # Use the app_logger to log every non-404 request before each request
     # See https://flask.palletsprojects.com/en/2.2.x/api/#flask.Flask.before_request
     app.before_request(lambda: _log_route(app_logger))
 
     app_logger.info("initialized flask logger")
 
 
+def add_extra_data_to_current_request_logs(data: dict[str, str]) -> None:
```
chouinar:
```suggestion
def add_extra_data_to_current_request_logs(data: dict[str, str | int | float]) -> None:
```
not every project will necessarily use the same tool for metrics, but allowing additional types (or just doing Any and not worrying) might be better. 

A quick check for NewRelic, I know str/int/float all work - I'd assume these should all be acceptable.

lorenyu: hmm good point. i actually think we'll want to support any JSON primitive including bool and None/null type as well. i added your change and added boolean + None too

chouinar (approved):
LGTM!

PR #97, opened 2023-01-21
Refactor logging to extract complex functionality from log formatters

### LINKED ISSUE #87: Document and enhance standardized logging practices
Haven't had a chance to look at the logging behavior in the template, so would be good to look through it and document how it works and how it's used, and in the process clean up and add useful enhancements

One enhancement that was discussed was the ability to register parameters on the logger in a request context (see this [PR thread](https://github.com/navapbc/template-application-flask/pull/86#discussion_r1081410116)), which I created as a separate ticket https://github.com/navapbc/template-application-flask/issues/95

also consider looking into other ways of adding context info: https://docs.python.org/3/howto/logging-cookbook.html#context-info such as LoggingAdapter, Filters, etc

### LINKED ISSUE #91: Simplify logging config
Not sure since I don't have historical context for the logging config code, but it seems that we can configure a single console root logger, and just have filters for the individual package/module loggers for filtering out logs of a certain level, rather than having multiple loggers defined.

lorenyu:
## Ticket

Resolves #91 
Resolves #87

## Changes
* Extract functionality of adding Flask request context data out of log formatters
  into flask_logger module
* Extract PII masking functionality to pii module
* Move mask_pii and mask_pii_for_key from string_utils to pii module
* Use logging module main API for configuration instead of logging.config.dictConfig
* Remove custom implementation of log capture fixture
* Import standard logging module instead of api.logging across application
* Automatically log request at the beginning of each request
* Bring test coverage to 100% for updated modules

## Context for reviewers

As I was looking at the logging code to document how it works (ticket #87) and add the enhancement to add custom data to current request's logs (ticket #95), I observed that a lot of functionality is mixed into the formatter classes. In particular, the log formatters weren't just doing formatting, but they were also (1) adding additional attributes from the flask request context, and (2) masking PII. This makes the logging system much harder to maintain, since complexity of the formatters is so high as they are doing so many things. It also creates inconsistencies and duplicate code. For example, the logic to add request context data is duplicated in both the JSON and human readable formatter, and would need to be duplicated for any new formatter. As another example, PII masking is only done for the JSON formatter, not the human readable formatter, which is an inconsistency that isn't obvious from looking at the high level purpose of the two formatters.

This looks like a big PR but there's no new functionality in this PR. A lot of the new lines are docstrings and comments, and the boilerplate imports and whitespace you get when separating out functionality into multiple files. The following summarizes the changes:

1. This PR pulls out the logic that isn't related to formatting out of the log formatters, which simplifies the code to be more maintainable as each module's functionality is now more restricted.

2. This PR also aims to reduce the number of generic "utils" by moving mask_pii and mask_pii_for_key to the api.logging.pii module rather than leaving it in a generic string_utils module.

3. As part of this refactoring, I also noticed a fix_logging pytest fixture and a lot of code in test_logging that seems to emulate pytest's caplog. I realized that it's because the way the application configures logging uses logging.config.dictConfig which blows away any existing logging handlers. In particular, it clears the caplog handler that the caplog fixture adds to the root logger. This PR changes the configuration to use the main logging API which involve non-destructive functions like addHandler, which more cleanly ignores any existing handlers. This is more appropriately follows the listener pattern and also allows us to remove a lot of extra testing code.

4. Finally, I didn't really understand the purpose of forcing the application to include api.logging instead of the standard module logging, and to rename getLogger as get_logger. The benefit doesn't seem to outweigh the cons of reducing the portability/reusability of code and the necessity of learning a slightly different way of logging. Since logging is a core part of Python, I refactored the application to use the standard library, and only use api.logging for initialization and configuration and custom enhanced behavior.

With this refactor, we are now more easily able to add additional enhancements while keeping complexity low. In this PR we added the ability to automatically log the request at the beginning of each flask request, so we no longer need to do that in each route. And in a follow up PR https://github.com/navapbc/template-application-flask/pull/96 we are able to add a function that adds custom data that gets added to all logs for the current flask request.

## Testing
CI, with 100% test coverage for the api.logging package (not including api.logging.audit)


app/api/logging/config.py
```diff
@@ -0,0 +1,49 @@
...
+    """Return the formatter used by the root logger.
+
+    The formatter is determined by the environment variable LOG_FORMAT. If the
+    environment variable is not set, the JSON formatter is used by default.
+    """
+
+    log_format = os.getenv("LOG_FORMAT", "json")
+
+    print("get_formatter")
+    print(log_format)
```
chouinar:
```suggestion

```

chouinar (commented):
I like the restructuring, I think the code-defined config is a lot easier to read/understand than the dictionary.

I did test this out locally real quick, and using the swagger endpoint, I'm not seeing any logs while running `make run-logs`. Some initial logs appeared for the audit log, but nothing for the API itself.

Also, a bit tangential, but I see one of the audit logs writes all of the env vars to the logs which seems like a security problem:

```
main-app  | 17:19:51.748  api.logging.audit                    audit_log                    AUDIT    subprocess.Popen                                                                 audit.event_name=subprocess.Popen count=1 audit.args.0=/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.10/bin/python audit.args.1=['/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.10/bin/python', '-m', 'api'] audit.args.3={'PATH': '/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.10/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin', 'HOSTNAME': '66279ad93211', 'LOG_FORMAT': 'human-readable', 'POSTGRES_DB': 'main-db', 'ENVIRONMENT': 'local', 'API_AUTH_TOKEN': 'LOCAL_AUTH_12345678', 'PORT': '8080', 'AWS_ACCESS_KEY_ID': 'DO_NOT_SET_HERE', 'FLASK_APP': 'api.app:create_app', 'AWS_SECRET_ACCESS_KEY': 'DO_NOT_SET_HERE', 'HIDE_SQL_PARAMETER_LOGS': 'TRUE', 'DB_HOST': 'main-db', 'AWS_DEFAULT_REGION': 'us-east-1', 'PYTHONPATH': '/app/', 'POSTGRES_PASSWORD': 'secret123', 'POSTGRES_USER': 'local_db_user', 'USER_ROLE_CSV_OUTPUT_PATH': './', 'LANG': 'C.UTF-8', 'GPG_KEY': 'A035C8C19219BA821ECEA86B64E628F8D684696D', 'PYTHON_VERSION': '3.10.8', 'PYTHON_PIP_VERSION': '22.2.2', 'PYTHON_SETUPTOOLS_VERSION': '63.2.0', 'PYTHON_GET_PIP_URL': 'https://github.com/pypa/get-pip/raw/5eaac1050023df1f5c98b173b248c260023f2278/public/get-pip.py', 'PYTHON_GET_PIP_SHA256': '5aefe6ade911d997af080b315ebcb7f882212d070465df544e1175ac2be519b4', 'HOST': '0.0.0.0', 'HOME': '/root', 'PLAT': 'linux-x86_64', 'VIRTUAL_ENV': '/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.10', 'WERKZEUG_SERVER_FD': '4', 'WERKZEUG_RUN_MAIN': 'true'}
main-app  | 17:19:51.750  api.logging.audit                    audit_log                    AUDIT    os.posix_spawn                                                                   audit.event_name=os.posix_spawn count=1 audit.args.0=/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.10/bin/python audit.args.1=['/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.10/bin/python', '-m', 'api'] audit.args.2={'PATH': '/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.10/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin', 'HOSTNAME': '66279ad93211', 'LOG_FORMAT': 'human-readable', 'POSTGRES_DB': 'main-db', 'ENVIRONMENT': 'local', 'API_AUTH_TOKEN': 'LOCAL_AUTH_12345678', 'PORT': '8080', 'AWS_ACCESS_KEY_ID': 'DO_NOT_SET_HERE', 'FLASK_APP': 'api.app:create_app', 'AWS_SECRET_ACCESS_KEY': 'DO_NOT_SET_HERE', 'HIDE_SQL_PARAMETER_LOGS': 'TRUE', 'DB_HOST': 'main-db', 'AWS_DEFAULT_REGION': 'us-east-1', 'PYTHONPATH': '/app/', 'POSTGRES_PASSWORD': 'secret123', 'POSTGRES_USER': 'local_db_user', 'USER_ROLE_CSV_OUTPUT_PATH': './', 'LANG': 'C.UTF-8', 'GPG_KEY': 'A035C8C19219BA821ECEA86B64E628F8D684696D', 'PYTHON_VERSION': '3.10.8', 'PYTHON_PIP_VERSION': '22.2.2', 'PYTHON_SETUPTOOLS_VERSION': '63.2.0', 'PYTHON_GET_PIP_URL': 'https://github.com/pypa/get-pip/raw/5eaac1050023df1f5c98b173b248c260023f2278/public/get-pip.py', 'PYTHON_GET_PIP_SHA256': '5aefe6ade911d997af080b315ebcb7f882212d070465df544e1175ac2be519b4', 'HOST': '0.0.0.0', 'HOME': '/root', 'PLAT': 'linux-x86_64', 'VIRTUAL_ENV': '/root/.cache/pypoetry/virtualenvs/template-application-flask-o9msT97p-py3.10', 'WERKZEUG_SERVER_FD': '4', 'WERKZEUG_RUN_MAIN': 'true'}
```

app/api/logging/__init__.py
```diff
@@ -82,13 +69,4 @@ def init(program_name: str) -> None:
             ),
         },
     )
-    logger.info("invoked as: %s", " ".join(original_argv))
-
-
-def get_logger(name: str) -> logging.Logger:
-    """Return a logger with the specified name."""
-    return logging.getLogger(name)
```
jamesbursa, line 90: For context, the original implementation had an additional check to try to encourage correct usage. It was like this:

```python
    if not name.startswith("massgov."):
        raise ValueError(
            "invalid logger name %r (try passing __package__ instead of __name__)" % name
        )
```

app/api/logging/config.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Loggers can be configured using config functions defined
+    # in logging.config or by directly making calls to the main API
+    # of the logging module (see https://docs.python.org/3/library/logging.config.html)
+    # We opt to use the main API using functions like `addHandler` which is
+    # non-destructive, i.e. it does not overwrite any existing handlers.
+    # In contrast, logging.config.dictConfig() would overwrite any existing loggers.
+    # This is important during testing, since fixtures like `caplog` add handlers that would
+    # get overwritten if we call logging.config.dictConfig() during the scope of the test.
+    consoleHandler = logging.StreamHandler(sys.stdout)
```
jamesbursa: Nit: snake case for local vars: `console_handler`

app/api/logging/flask_logger.py
```diff
@@ -0,0 +1,108 @@
...
+    # messages are passed directly to the ancestor loggers’ handlers bypassing any filters
+    # set on the ancestors.
+    # See https://docs.python.org/3/library/logging.html#logging.Logger.propagate
+    for handler in app_logger.handlers:
+        handler.addFilter(_add_app_context_info_to_log_record)
+        handler.addFilter(_add_request_context_info_to_log_record)
+
+    # Use the app_logger to log every non-404 request before each request
+    # See https://flask.palletsprojects.com/en/2.2.x/api/#flask.Flask.before_request
+    app.before_request(lambda: _log_route(app_logger))
```
jamesbursa, line 52: Do we want to log at the end of the request too? Where we can log things like response code, size or response, duration.

app/api/logging/pii.py
```diff
@@ -0,0 +1,102 @@
...
+    logger.addFilter(pii.mask_pii)
+"""
+
+import logging
+import re
+from typing import Any, Optional
+
+# Attributes of LogRecord to exclude from the JSON formatted lines. An exclusion list approach is
+# used so that all "extra" attributes can be included in a line.
+EXCLUDE_ATTRIBUTES = {
```
jamesbursa: Do we want to filter by this list again here? This is the same list as in formatters.py and is more about what is output than where to mask pii.

app/tests/api/logging/test_formatters.py
```diff
@@ -0,0 +1,44 @@
...
+        "name": "test_json_formatter",
+        "msg": "hello",
+        "levelname": "WARNING",
+        "levelno": 30,
+        "filename": "test_formatters.py",
+        "module": "test_formatters",
+        "funcName": "test_json_formatter",
+        "foo": "bar",
+    }
+    assert json_record | expected == json_record
```
jamesbursa: What's the advantage over doing `json_record == expected`?

app/tests/api/util/test_string_utils.py
```diff
@@ -19,37 +19,3 @@ def test_blank_for_null():
     assert blank_for_null("hello") == "hello"
     assert blank_for_null(4) == "4"
     assert blank_for_null(["a", "b"]) == "['a', 'b']"
-
-
-def test_mask_pii():
```
jamesbursa, line 24: Did this test get lost when the function was moved? I think it's valuable as the regexp `TIN_RE` is complex.

jamesbursa (commented):
Thanks for making all these improvements. Only minor comments.

app/api/logging/__init__.py
lorenyu: thanks for the context! with the handler now attached to the root logger, that's no longer necessary i assume?

app/api/logging/config.py
lorenyu: good catch

app/api/logging/flask_logger.py
lorenyu: Yeah I think that's a good idea, but since that's new functionality (we weren't doing that before), I'll create a separate ticket for that to keep this PR as a refactor PR: https://github.com/navapbc/template-application-flask/issues/99

app/api/logging/pii.py
lorenyu: Thanks for the context and for noticing the duplication, I wasn't aware of that. I'll remove it from here!

lorenyu: Oh hm I notice I'm actually not using that dictionary anywhere if I remove it from here. Sounds like the intention is to exclude those fields? Is it because it's too noisy?

app/tests/api/logging/test_formatters.py
lorenyu: it's because the the log record contains a lot of other data that i don't really care about, like line number, thread number etc. if i change it to json_record == expected i get the following diff

image

this is essentially the same as `_assert_dict_contains` in `test_flask_logger.py`

should i just add a comment here explaining it? or maybe create a `_assert_dict_contains` test utility?

app/tests/api/util/test_string_utils.py
lorenyu: probably, good catch i'll bring it back, sorry for losing it!

lorenyu:
> I did test this out locally real quick, and using the swagger endpoint, I'm not seeing any logs while running make run-logs. Some initial logs appeared for the audit log, but nothing for the API itself.

@chouinar weird, must've broken something in a later commit ... will take a look and see what happened

lorenyu:
@chouinar looks like when I switched over the config format to use the main API in b33073e58bfb414cccb3fe5340bf07590c944091 I forgot to set log level to INFO so it was only logging warnings and above

app/tests/api/util/test_string_utils.py
lorenyu: Good catch, I actually introduced a bug that I wouldn't have caught if I didn't bring this test back. Take another look to see if it looks right. It's in a new file called test_pii.py

lorenyu: One small change in behavior to note, I didn't end up casting the input to a string if there wasn't a regex match. It broke things to change types of stuff in the record that log handlers weren't expecting to have changed, like `record.created`.

lorenyu:
@chouinar address the env var logging in a different PR https://github.com/navapbc/template-application-flask/pull/101

app/tests/api/logging/test_formatters.py
lorenyu: I moved it to a util `tests.lib.assertions`

app/api/logging/pii.py
lorenyu: i tried logging a few things in healthcheck as a test and don't see any random attributes from that list even when i completely remove this dictionary. can't seem to figure out what it does.

image

i just went ahead and removed it. also, even if we do end up seeing logs with filename, lineno, etc. i don't see much of an issue with that being in the logs

chouinar:
> @chouinar address the env var logging in a different PR #101

Is that going to be merged into this one? I can't review this further without testing it a bit

chouinar:
@lorenyu - I'm having a hard time following the PRs for this work as it's split across 5 or 6 active PRs that fix issues in this, however I'm hesitant to approve anything with active issues as I've already seen one team make a copy of the template repo while you were actively modifying the DB session logic. I think if the template repo is able to be copied, it needs to start having releases to make it safe to make changes like this, otherwise someone might pull in half-working logging

lorenyu:
@chouinar sorry I realize there are a lot of PRs.

#101 is a separate PR that fixes the issue with logging env vars that you discovered, which is already an issue on `main` so that's why that's a separate PR and not in this one

otherwise this PR should be standalone and should be have fully working logging i.e. there shouldn't be anything in this PR that is half-working.

the other PRs #94 (log query params) and #96 (function for adding extra data to current request) are enhancements that i was able to do after this PR. i set the base branch for those PRs to this branch to make the diff easier, but i don't want to merge those until we merge this one so that they can live as separate commits.

app/api/logging/pii.py
jamesbursa: I think you're not seeing them because there's also similar code in decodelog that excludes them from the extra part of the output (blue part). So I agree with this change.

jamesbursa (approved):

lorenyu:
@chouinar I saw James just approved the PR. Should I wait for your review as well or do you want me to go ahead and merge?

chouinar:
> @chouinar I saw James just approved the PR. Should I wait for your review as well or do you want me to go ahead and merge?

I haven't had a chance to get another look at this, and probably wouldn't be able to until tomorrow or early next week. A quick skim doesn't raise any immediate concerns.

lorenyu:
@chouinar oh ok no worries and no rush! I'll go ahead and merge then for now, and I'd be happy to continue to address anything that comes up if you take a look later.

PR #103, opened 2023-01-24
Define logging conventions

lorenyu:
## Ticket

Related to #87 

## Changes
see title

## Context for reviewers
Define some logging best practices that have been useful across multiple projects

## Testing
n/a

### DIFF
--- a/docs/app/monitoring-and-observability/logging-conventions.md
+++ b/docs/app/monitoring-and-observability/logging-conventions.md
@@ -1,0 +1,48 @@
+# Logging Conventions
+
+## Purpose
+
+Logging is a valuable tool for engineering teams to support products in production. It enables debugging, real-time alerts, and real-time business metrics. This document is for engineers to have a set of guidelines and rationale for how to approach logging events in our system.
+
+## Principles
+
+### Make code observability a primary tool for debugging and reasoning about production code
+
+When a user runs into an issue in production, logs offer one of the primary ways of understanding what happened. This is especially important for situations where we can’t or don’t know how to reproduce the issue. In general it is not feasible to attach a debugger to production systems, or to set breakpoints and inspect the state of the application in production, so logs offer a way to debug through “print statements”.
+
+### Make it easy for on-call engineers to search for logs in the codebase
+
+While it is important to use logs to reason about the code, it is often equally important to be able to understand what code is being run from the logs. This is especially true for on-call engineers who may encounter production issues that are in areas of the codebase that they are not directly familiar with. Thus it is important to make it easy to find the associated place in the code that triggered logs.
+
+### Optimize for ease, flexibility, and speed of queries over DRY principles
+
+Log querying systems are often limited in their querying abilities. Most log databases are not relational databases, and as such to enable effective querying of logs, we should think about how we design the structure of our log messages and metadata.
+
+## Guidelines
+
+### General
+
+- **Minimize logic or dynamic behavior** – Logs are often used for troubleshooting, so minimize dynamic behavior in logs to reduce the chance that the logging system itself could be contributing to adding complexity to an issue.
+- **Log all key points throughout the application** – Include logs for:
+  - Important branches in business logic (so you can reason from the logs which branch the code took in production without the aid of a debugger)
+  - Before and after actions that are important or actions that have a nontrivial chance of failing (e.g. external service calls, or calls to legacy systems)
+- **Log both successes and failures** – Successes can be useful for debugging what steps were completed successfully to narrow down what failed. In addition, logging successes can be useful for powering real-time business metrics.
+
+### Log event type
+
+- **INFO for user/client error** – Use INFO events if it’s a user or client error, and ERROR events if it’s a system error. This helps us set stricter alerts on system errors that won’t be affected by fluctuations in user behavior.
+
+### Log messages
+
+- **Standardized log messages** – Consistently formatted and worded log messages easier to read when viewing many logs at a time, which reduces the chance for human error when interpreting logs. It also makes it easier to write queries by enabling engineers to guess queries and allow New Relic autocomplete to show available log message options to filter by.
+- **Statically defined log messages** – Avoid putting dynamic data in log messages. Static messages are easier to search for in the codebase. Static messages are also easier to query for those specific log events without needing to resort to RLIKE queries with regular expressions or LIKE queries.
+
+### Attributes
+
+- **Log primitives not objects** – Explicitly list which attributes you are logging to avoid unintentionally logging PII. This also makes it easier for engineers to know what attributes are available for querying, or for engineers to search for parts of the codebase that logs these attributes.
+- **Structured metadata in custom attributes** – Put metadata in custom attributes (not in the log message) so that it can be used in queries more easily. This is especially helpful when the attributes are used in "group by" clauses to avoid needing to use more complicated queries.
+  - **system identifiers** – Log all relevant system identifiers (uuids, foreign keys)
+  - **correlation ids** – Log ids that can be shared between frontend events, backend logs, and ideally even sent to external services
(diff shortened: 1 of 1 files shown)


docs/app/monitoring-and-observability/logging-conventions.md
```diff
@@ -0,0 +1,48 @@
...
+
+- **Minimize logic or dynamic behavior** – Logs are often used for troubleshooting, so minimize dynamic behavior in logs to reduce the chance that the logging system itself could be contributing to adding complexity to an issue.
+- **Log all key points throughout the application** – Include logs for:
+  - Important branches in business logic (so you can reason from the logs which branch the code took in production without the aid of a debugger)
+  - Before and after actions that are important or actions that have a nontrivial chance of failing (e.g. external service calls, or calls to legacy systems)
+- **Log both successes and failures** – Successes can be useful for debugging what steps were completed successfully to narrow down what failed. In addition, logging successes can be useful for powering real-time business metrics.
+
+### Log event type
+
+- **INFO for user/client error** – Use INFO events if it’s a user or client error, and ERROR events if it’s a system error. This helps us set stricter alerts on system errors that won’t be affected by fluctuations in user behavior.
```
KevinJBoyer: I don't know if it's possible to do this for a template, but some guidance on when to use info vs. warning vs error generally would be useful.

E.g., for the payments system on PFML:
 - `warning` has no impact that (vs. `info`)
 - but `error` triggers a low priority alert for the on-call engineer to investigate

That's behavior, so maybe that's project specific, but we've had some back and forth on the team of what the semantics should be. My take is:
 - INFO -- nothing's wrong with the system (though as noted here, the user may have done something erroneously.)
 - WARNING -- something _may_ be wrong with the system, a human should look into this but it's not urgent.
 - ERROR -- something has _definitely_ gone wrong with the system and require human intervention.
 
 But others would say `warning` should be treated as "somewhat out of the ordinary but not necessarily erroneous behavior" that only needs investigation if there are _more_ warnings than usual on a given day. (E.g., usually some process logs 25 warnings, yesterday it logged 200.)

docs/app/monitoring-and-observability/logging-conventions.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Log messages
+
+- **Standardized log messages** – Consistently formatted and worded log messages easier to read when viewing many logs at a time, which reduces the chance for human error when interpreting logs. It also makes it easier to write queries by enabling engineers to guess queries and allow New Relic autocomplete to show available log message options to filter by.
+- **Statically defined log messages** – Avoid putting dynamic data in log messages. Static messages are easier to search for in the codebase. Static messages are also easier to query for those specific log events without needing to resort to RLIKE queries with regular expressions or LIKE queries.
+
+### Attributes
+
+- **Log primitives not objects** – Explicitly list which attributes you are logging to avoid unintentionally logging PII. This also makes it easier for engineers to know what attributes are available for querying, or for engineers to search for parts of the codebase that logs these attributes.
```
okkays, line 44: It might be worth defining a standard or utility for performance-related log messages, if you don't already have one. For example: https://stackoverflow.com/a/34981726

I've found it useful to have a decorator to surface performance issues as info/warning/error based on the time it takes - e.g.:

```py
@log_performance(warning_after=2, error_after=10)
def do_complicated_thing():
  for every_record in my_database:
    do_something_expensive()
```

lorenyu: Good idea! Created a backlog item for that idea so we don't forget it: https://github.com/navapbc/template-application-flask/issues/105

lorenyu: @okkays thinking about this more, isn't this something that monitoring tools like newrelic already provide out of the box? i.e. tracking request times, being able to alert on request times, etc

docs/app/monitoring-and-observability/logging-conventions.md
lorenyu: @KevinJBoyer thanks for calling out the lack of guidance on when to use warning vs error! agree with the semantics you suggested. i updated this section can you take another look

KevinJBoyer: Nice, yeah this makes a ton of sense to me and helps clear up the ambiguity!

KevinJBoyer (approved):
Not sure if you're looking for formal approval on these PRs from those of us poking our heads in, but just in case :-)

docs/app/monitoring-and-observability/logging-conventions.md
okkays: Sorry for the late reply!

Request times are like end-to-end tests in this case - this lets us track performance of each unit.

A controller route might look like, for example:

```python
@app.route('/report/countries', methods=['GET'])
def get_country_split():
  check_access()  # fetches/checks a user object from an NDB store
  raw = get_records(flask.request.get('filters')) # "simple" fetch from database
  grouped = sum_by_country(raw)  # hacky python grouping that should be rewritten in orm/sql
  return format_as_json(grouped)
```

If that starts getting slow, it could be for many reasons, e.g.:
- our permissions check is timing out because ndb is disorganized or wonky
- the "simple" database fetch is complicated by ambitious filter overengineering
- the sum_by_country is doing things in python that should be done in the data layer
- our json formatter is recursively serializing extra metadata that we don't mean it to

This would let us quickly see whether the access overhead, fetching, transformation, or formatting/response crafting is the problem.

I've personally found this to be especially useful debugging tricky NDB or migration operations, or in cases where I want to prove that we're asking the system to do something it wasn't designed for, and we need to set up more efficient views or refactor a particular part. You can manually get similar analysis by logging after each action, but you can't escalate individual components to warnings/failures based on how far apart the log entries are

lorenyu: No worries about the delay!

To make sure I understand, is the idea that you'd add the `@log_performance` decorator to each of the inner functions?

And is the idea to have tests in the automated test suite that ensures functions run under a certain time (asking since you mention that this was like e2e tests)? Or more that you'd add production monitoring for those warning logs that get emitted when a function takes a long time, and alert when those warning logs are detected?

okkays: > To make sure I understand, is the idea that you'd add the @log_performance decorator to each of the inner functions?

Yep - the decorator works great for active debugging, or for keeping an eye on functions we know might be problematic. Another way it could be implemented, that works great for tracking a request across a controller or problematic function, is to have the decorator pass a timer to the function:

```py
@app.route('/report/countries', methods=['GET'])
@timed(warning=2, error=5)
def get_country_split(timer):
  check_access()  # fetches/checks a user object from an NDB store
  timer.next('Permissions checked!')

  raw = get_records(flask.request.get('filters')) # "simple" fetch from database
  timer.next('Fetched from DB!')

  grouped = sum_by_country(raw)  # hacky python grouping that should be rewritten in orm/sql
  timer.next('Grouped!')

  return format_as_json(grouped)
```

which would log something like:

```
[INFO ] [timestamp] step 0.0s | total 0.0s - get_country_split: Start
[INFO ] [timestamp] step 1.0s | total 1.0s - get_country_split: Permissions Checked!
[WARN ] [timestamp] step 3.0s | total 4.0s - get_country_split: Fetched from DB!
[ERROR] [timestamp] step 6.0s | total 10.0s - get_country_split: Grouped!
[ERROR] [timestamp] step 7.0s | total 17.0s - get_country_split: Finish
```

>  more that you'd add production monitoring for those warning logs that get emitted when a function takes a long time, and alert when those warning logs are detected?

This - you can change the log category based on how long the function takes to finish. They are better for monitoring bloat over time, as you can usually recognize with general QA when a brand new feature is not meeting performance goals


Happy Friday!

Issue #98, opened 2023-01-23
Look into alternative approach for configuration


lorenyu:
quick discussion with @jamesbursa on whether we should consider moving non-secret configs to config files

also look at https://flask.palletsprojects.com/en/2.2.x/config/

jamesbursa:
Some rough thoughts on this. Using environment variables for configuration, as currently, has some disadvantages:

- Changing the **app** configuration requires **infra** changes (e.g. in container definitions in terraform), which can mean cross-team coordination, sign-off etc.
- Embedding `os.getenv` calls or equivalent in the code is fragile - if the env var is missing it can have unexpected behaviour, only strings are supported etc.
- Unit tests become more fragile - they can change behaviour if env vars are present in the developers environment, monkeypatching is needed if target code uses getenv
- There are multiple ways & places to set them (sometimes hidden or not obvious) - in the shell, docker-compose.yml, terraform / AWS, developer's IDE config, monkeypatching in tests

jamesbursa:
A possible vision of how a better system could look.

## Configuration files
Somewhere is the codebase is a directory of configuration files, one per deployed environment (prod, stage, test, etc.), plus one for local development and one for the test suite (last two might be the same).

Each file is Python (or yaml or json), containing configuration in a simple format:
```python
db_host = "localhost"
db_schema = "api"
bucket_name = "api-bucket-test"
logging_level_override = {"api": "INFO", "sqlalchemy": "WARNING"}
...
```

These files get included in the container with the code. When the system starts, it loads the correct config depending on some detection of which environment it's running in. That could be an env var, or a required command line flag, with the name or path of the config file.

Questions:

- Should it be more structured, typed, and hierarchical?
- Some way to have a default and only specify the differences?
- Some way to deploy new config independent of code? Or is it really an advantage to keep them in sync?

## Usage in the code
Two possible approaches:

- A central config module that other modules can retrieve the setting from.
- Something more module-based. Each module defines its own config setting class. It's read from the file when the program starts. Or in tests, defined with fixed values in the code.

PR #104, opened 2023-01-24
Move logging config to use pydantic base env config

### LINKED ISSUE #93: Use PydanticBaseConfig for logging config
logging config manually pulls from LOG_FORMAT env var, but we should use the same pattern that's used for DbConfig and use the PydanticBaseConfig that does this automatically

lorenyu:
## Ticket

Resolves #93 

## Changes
see title

## Context for reviewers
Just a small change to make our configs adopt a more consistent pattern of subclassing from PydanticBaseEnvConifg rather than manually reading from env vars.

There's [another ticket to revisit configs altogether](https://github.com/navapbc/template-application-flask/issues/98), but that's out of scope for this ticket

## Testing
CI


app/api/logging/config.py
```diff
@@ -1,9 +1,16 @@
...
 import api.logging.formatters as formatters
 import api.logging.pii as pii
+from api.util.env_config import PydanticBaseEnvConfig
+
+logger = logging.getLogger(__name__)
+
+
+class LoggingConfig(PydanticBaseEnvConfig):
+    log_format: str = "json"
+    log_level: int = logging.INFO
```
jamesbursa, line 13: Would it be easier for the user to make this a str instead? `setLevel()` accepts a string too. It will raise an error if it's invalid so that would happen in `configure_logging()`.

```python
>>> import logging
>>> logging.root.setLevel("INFO")
>>> logging.root.setLevel("DEBUG")
>>> logging.root.setLevel("NOTE")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.9/logging/__init__.py", line 1417, in setLevel
    self.level = _checkLevel(level)
  File "/usr/lib/python3.9/logging/__init__.py", line 194, in _checkLevel
    raise ValueError("Unknown level: %r" % level)
ValueError: Unknown level: 'NOTE'
```

jamesbursa (approved):

app/api/logging/config.py
lorenyu: I see, so do something like `log_level: str = "INFO"`? Sure that does seem easier than the int version.

PR #108, opened 2023-01-28
Log request life cycle

### LINKED ISSUE #99: Log standard message at the end of every request
Idea from James to log things like response code, size or response, duration at the end of every request https://github.com/navapbc/template-application-flask/pull/97#discussion_r1084457234

lorenyu:
## Ticket

Resolves #99 

## Changes
* Change start request log to have static message to follow logging-conventions.md
* Log request end
* Shorten human readable log message to 50 characters

## Context for reviewers
While reviewing #97, @jamesbursa had an idea of logging at the end of every request. This PR adds that functionality, and also changes the start request log to use a static message string to follow the logging conventions (which makes the structured logs easier to query using log query tools).

## Testing
Example logs
image


app/api/logging/flask_logger.py
```diff
@@ -73,17 +73,38 @@ def add_extra_data_to_current_request_logs(
...
+    logger.info(
+        "end request",
+        extra={
+            "response.status_code": response.status_code,
+            "response.content_length": response.content_length,
+            "response.content_type": response.content_type,
+            "response.mimetype": response.mimetype,
+        },
+    )
+    return response
```
chouinar, line 107: Might be tricky to do / not worthwhile, but what about a metric to calculate the request duration in milliseconds?

chouinar (approved):
LGTM! Just added an idea for another possible metric

app/api/logging/flask_logger.py
lorenyu: good idea! created a ticket for that https://github.com/navapbc/template-application-flask/issues/111

PR #110, opened 2023-01-28
update breaking internal links

lxqnt:
## Ticket
No ticket


lorenyu (approved):
Thanks for making these fixes!

btw when filling out the PR template, it's ok to remove the block quote formatting, so the text can be something

## Context for reviewers
did something

as opposed to

## Context for reviewers
> did something

PR #122, opened 2023-02-03
Fix tests that fail when run outside Docker

### LINKED ISSUE #119: Some tests fail when running outside container
🔒 [michael found this see thread](https://nava.slack.com/archives/C03G1SWD9H7/p1675439828424959?thread_ts=1675439825.157059&cid=C03G1SWD9H7)

introduced in PR #101

lorenyu:
## Ticket

Resolves #119 

## Changes
- Include db_client fixture which sets DB-related environment variables
  that aren't set by default when running outside the container

## Context for reviewers
Outside the container the environment variables from local.env aren't automatically set which caused some db tests to fail. This PR includes the db_client fixture that sets those environment variables.

## Testing
ran `poetry run pytest -m "not audit"` which now passes
also ran `make test`

### DIFF
--- a/.github/workflows/ci-app.yml
+++ b/.github/workflows/ci-app.yml
@@ -36,4 +36,7 @@ jobs:
         run: make lint-security
 
       - name: Start tests
-        run: make test-coverage
+        run: |
+          make test-audit
+          make test-coverage
+        
--- a/app/Makefile
+++ b/app/Makefile
@@ -164,7 +164,7 @@ lint-poetry-version: ## Check poetry version
 	grep --quiet 'lock-version = "1.1"' poetry.lock
 
 lint-security: # https://bandit.readthedocs.io/en/latest/index.html
-	$(PY_RUN_CMD) bandit -r . --number 3 --skip B101 -ll -x ./.venv
+	$(PY_RUN_CMD) bandit -c pyproject.toml -r . --number 3 --skip B101 -ll -x ./.venv
 
 
 ##################################################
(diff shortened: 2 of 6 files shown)


chouinar (commented):
I think these had previously worked for me because `load_local_env_vars` was automatically called before every test, but that was removed. Not sure if we want to consider re-adding that?

That or move the monkeypatch params set in `create_isolated_db` to their own fixture and either auto-set that, or make that get called here (and automatically on db_client)

chouinar:
Running locally, I still have 3 failures:
```
FAILED tests/api/route/test_user_route.py::test_create_user_bad_request[missing all required fields] - assert 400 == 422
FAILED tests/api/route/test_user_route.py::test_create_user_bad_request[invalid types] - assert 400 == 422
FAILED tests/api/route/test_user_route.py::test_create_user_bad_request[invalid role type] - assert 400 == 422
```

lorenyu:
@chouinar first responding to the remaining failures – you may have to run poetry install again. the new version of apiflask changes 400 to 422 for bad requests (see https://apiflask.com/changelog/#version-120)

yeah adding `load_local_env_vars` makes sense. should that be an autouse fixture with session scope or do you think we should set it on every test?

lorenyu:
Ok done! PTAL

chouinar:
> yeah adding `load_local_env_vars` makes sense. should that be an autouse fixture with session scope or do you think we should set it on every test?

I have it set as autouse in the NJ code: https://github.com/navapbc/archive-nj-dol-ui-claimant-api/blob/main/app/tests/conftest.py#L27 

Haven't seen any issues with it other than pycharm not working for one very specific test I haven't gotten around to looking into.

--

And yes, poetry install fixed those tests

chouinar (approved):
LGTM!

PR #126, opened 2023-02-04
Fix OpenAPI spec generation

### LINKED ISSUE #63: Keep openapi.yml file in sync with code
# Context

`make openapi-spec` will generate the openapi.yml file from the schema files, but this is currently manually run, so it can get out of sync

# Options

1. have a CI check that runs `make openapi-spec` and if it changes the file (e.g. the git index isn't clean) then fail the CI check
2. have a workflow that runs `make openapi-spec` on PRs and if there are any changes automatically commit and push the changes

### LINKED ISSUE #69: Add test or check for swagger endpoint
And add some tests too

lorenyu:
## Ticket

Resolves #63 
Resolves #69 

## Changes
* Fix make openapi-spec
* Make create_app factory work with no arguments
* Add check_db_connection_on_init DbConfig setting
* Remove now unneeded test_logger fixture
* Make tests more performant by making app fixture session scope
* Change openapi-spec to use --output CLI param instead of piping stdout to avoid logs showing up in spec 
* Add workflow to update openapi spec on application changes

## Context for reviewers
I noticed that `make openapi-spec` was broken for a while ever since I made the `create_app` factory require arguments. Upon re-reading [the Flask docs on the application factories pattern](https://flask.palletsprojects.com/en/2.2.x/patterns/appfactories/) I realized that the `create_app` factory should ideally work without any arguments. Arguments should at most be configuration overrides like `environment`. So this PR moves the extension object (logging, db) creation into create_app, and moves the "check db connection on init" functionality to a DbConfig setting that can be adjusted by environment variables. This simplifies test fixtures (we are able to remove the test_logger fixture).

All this has the result of fixing `make openapi-spec`.

Finally, to keep the openapi spec in sync going forward, I added a workflow that automatically updates openapi.yml when there are changes to the application's API.

## Testing
Ran


chouinar (approved):
LGTM!

PR #127, opened 2023-02-04
Remove redundant setting of db env vars

lorenyu:
## Ticket

n/a

## Changes
see title

## Context for reviewers
The PR #122 introduced an autouse fixture that loaded local.env during tests.
As a result, most of the env vars set while creating the isolated DB schema are no longer necessary.
This PR removes those redundant env vars and DRY's things up a bit, making the resulting function
easier to read and understand.

## Testing
Run `make test` and `poetry run pytest -m "not audit"` (running tests both inside and outside of Docker)

### DIFF
--- a/app/tests/lib/db_testing.py
+++ b/app/tests/lib/db_testing.py
@@ -18,10 +18,6 @@ def create_isolated_db(monkeypatch) -> None:
     """
     schema_name = f"test_schema_{uuid.uuid4().int}"
     monkeypatch.setenv("DB_SCHEMA", schema_name)
-    monkeypatch.setenv("POSTGRES_DB", "main-db")
-    monkeypatch.setenv("POSTGRES_USER", "local_db_user")
-    monkeypatch.setenv("POSTGRES_PASSWORD", "secret123")
-    monkeypatch.setenv("ENVIRONMENT", "local")
 
     # To improve test performance, don't check the database connection
     # when initializing the DB client.


app/tests/lib/db_testing.py
```diff
@@ -18,10 +18,6 @@ def create_isolated_db(monkeypatch) -> None:
     """
     schema_name = f"test_schema_{uuid.uuid4().int}"
     monkeypatch.setenv("DB_SCHEMA", schema_name)
-    monkeypatch.setenv("POSTGRES_DB", "main-db")
-    monkeypatch.setenv("POSTGRES_USER", "local_db_user")
-    monkeypatch.setenv("POSTGRES_PASSWORD", "secret123")
-    monkeypatch.setenv("ENVIRONMENT", "local")
```
chouinar, line 24: No issue with removing these, but I've seen developers wanting to test things locally with environment+DB values set to something that is not `local` before, and forgetting to undo that before running tests (i've done it at least once).

Do we want to add any sort of assertion that you aren't trying to run test with `environment` or the DB set to something they shouldn't be?

lorenyu: oh interesting, that's a good concern.
this makes me think we should have a test.env separate from local.env, similar to what we were talking about over slack. it'll also let us set LOG_LEVEL to "warning" or "critical" for test.env, which would alleviate the issue with noisy logs during test runs. thoughts?

lorenyu: @chouinar does Alvin's PR for handling dotenv vars solve our issues? would you be able to port the pattern over?

chouinar: I think I need to see that PR move further along to know for sure. Will let you know when it's settled

sawyerh: @chouinar Just checking if there was a conclusion to share back here

lorenyu: This same change was done in https://github.com/navapbc/template-application-flask/pull/188, so this PR is now obsolete. I'll close

lorenyu:
Made obsolete by https://github.com/navapbc/template-application-flask/pull/188

PR #132, opened 2023-02-07
Improve readability of logs

### LINKED ISSUE #116: Turn off security audit logging locally by default
The audit logs aren't generally useful during local development, so we should find a way to turn them off by default

### LINKED ISSUE #131: Add screenshots of human readable logging to docs
see https://github.com/navapbc/template-application-flask/pull/128#discussion_r1098754103

lorenyu:
## Ticket

Resolves #131 
Resolves #116 

## Changes
* Make log message width configurable
* Change default message width to 50 (from 80)
* Add screenshot of human readable logs to docs
* Shorten LoggingConfig variable names

## Context for reviewers
Made some of the improvements we talked about with the exception of removing func_name since that was a controversial, and added a screenshot

----

🔒 [slack thread](https://nava.slack.com/archives/C03G1SWD9H7/p1675788286728269)


app/api/logging/decodelog.py
```diff
@@ -72,20 +74,21 @@ def format_line(
...
-        level,
-        message,
-        BLUE,
-        format_extra(extra),
-        RESET,
+    return "{created}  {logger_color}{logger_name:<36}{reset_color} {func_name:<28} {level_color}{level:<8} {message} {extra_color}{extra}{reset_color}".format(
+        created=format_datetime(created),
+        logger_color=colour_for_name(name),
+        logger_name=name,
+        reset_color=RESET,
```
chouinar, line 84: Not for this PR, just thinking. Since the way color is implemented is more or less `<START COLOR> txt <STOP COLOR>` would it be clearer here to instead have the color functions more or less do that?

Something like:
```
def colorized_text(text: str, color: str) -> str:
      return f"{color}{text}{RESET}"
```
The width would make that a bit trickier but probably workable.

app/api/logging/config.py
```diff
@@ -1,16 +1,27 @@
...
+    message_width: int = formatters.HUMAN_READABLE_FORMATTER_DEFAULT_MESSAGE_WIDTH
+
+
 class LoggingConfig(PydanticBaseEnvConfig):
-    log_format: str = "json"
-    log_level: str = "INFO"
+    format = "json"
+    level = "INFO"
+    enable_audit = True
+    human_readable_formatter = HumanReadableFormatterConfig()
```
chouinar, line 20: How would I set something different here?

Is there some way to specify a path in the env var?

chouinar (approved):
LGTM, just one question, and a random spin-off idea

app/api/logging/decodelog.py
lorenyu: Yeah I had a similar thought! I think that would make the code a bit easier to read too.

app/api/logging/config.py
lorenyu: yeah I gave a commented out example in `local.env`, but you can do
`LOG_HUMAN_READABLE_FORMATTER__MESSAGE_WIDTH=80` to overwrite the values in `HumanReadableFormatterConfig`

PR #134, opened 2023-02-08
Issue-111: Logging request duration

### LINKED ISSUE #111: Log request duration
Idea from [this PR comment](https://github.com/navapbc/template-application-flask/pull/108#discussion_r1090947996) to log the request duration in milliseconds

implementation note:
- could potentially put start time on flask.g in before_request, then in after_request find difference between start and end time

rsk2:
## Ticket

Resolves #111 

## Changes
* Log request duration


app/api/logging/flask_logger.py
```diff
@@ -73,6 +74,11 @@ def add_extra_data_to_current_request_logs(
     setattr(flask.g, EXTRA_LOG_DATA_ATTR, extra_log_data)
 
 
+def _track_request_start_time() -> None:
+    """Store the request start time in flask.g"""
+    flask.g.request_start_time = datetime.now()
```
lorenyu: Let's use python's [time.perf_counter](https://docs.python.org/3/library/time.html#time.perf_counter) method which is more intended to use for performance measurements

app/api/logging/flask_logger.py
```diff
@@ -102,6 +108,7 @@ def _log_end_request(response: flask.Response) -> flask.Response:
             "response.content_length": response.content_length,
             "response.content_type": response.content_type,
             "response.mimetype": response.mimetype,
+            "duration": datetime.now() - flask.g.request_start_time
```
lorenyu: nit: i think API response times are most useful in milliseconds rather than seconds, so maybe let's report milliseconds elapsed and name 

nit: could we name this "response.time_ms" instead of "duration"?

lorenyu (commented):
thanks for the PR! left some comments

app/api/logging/flask_logger.py
rsk2: Changed

app/api/logging/flask_logger.py
rsk2: Changed

lorenyu (approved):
Looks good!

lorenyu (commented):
Oh actually looks like there are some lint errors and stuff, can you fix those? You can `make check` to see what errors you have locally.

rsk2:
@lorenyu I have fixed the errors

lorenyu:
Awesome!

PR #135, opened 2023-02-08
Move LRUDict to audit module

### LINKED ISSUE #121: Move LeastRecentlyUsedDict into the logging/audit module
[see thread](https://github.com/navapbc/template-application-flask/pull/101#discussion_r1096052457)

lorenyu:
## Ticket

Resolves #121 

## Changes
* Move LeastRecentlyUsedDict to audit module

## Context for reviewers
Moving the LeastRecentlyUsedDict to the audit module since it's only used in one place, and the particular implementation of the dict is specific to the audit since it is designed to store integers (it defaults values to `0`). We also want to minimize catch all utilities packages/modules to have cleaner dependencies and better portability.

## Testing
CI


chouinar (approved):
LGTM!

Issue #136, opened 2023-02-09
Consider a different way to load local environment vars


lorenyu:
The load_local_env_vars module seems unnecessary and is weird that we call that function in prod (it's a no-op). It creates a situation where local code branch is different from prod code based on environment variables.

## One idea

If we rename `local.env` to dotenv then it will automatically load.

Pros:
* This will also help remove code branches that differ between local and prod systems.
* This will allow local developers to have their personal config changes without accidentally committing it and messing up other people's configs. Shared configs can still go in something like `.env.template`.

Cons:
* The con is that new config changes will need to be communicated to devs so they can pull new config changes from .env.template.

## More context

Also see this PR discussion https://github.com/navapbc/template-application-flask/pull/138/files#r1105922982

lorenyu:
Another possibility for running on local host machine would be to let developers rely on tools that appropriately set environment variables in the shell. direnv is a good tool for this. We can have developers install direnv and then have an .envrc file that sources the local.env file.

PR #138, opened 2023-02-13
Use Flask custom CLI commands for custom scripts

### LINKED ISSUE #90: Fix background job script and consider using flask custom cli commands
i probably broke the background job script create_user_csv since haven't tested it in a while. consider using flask custom cli commands instead of our own manual entry point

lorenyu:
## Ticket

Resolves #90 

## Changes
* Use Flask custom CLI commands for custom scripts
* Remove main() entry point from create_user_csv service
* Remove script_utils
* Rename api.route package to api.api, and organize all things related to blueprints together:
   * X_blueprint.py
   * X_commands.py
   * X_routes.py
   * X_schemas.py
* Rewrite test_create_user_csv to use test_cli_runner rather than only test the service
* Add migration to cascade on delete
* Rewrite test_create_user_csv to truncate the user table rather than requiring an isolated db schema
* Remove now unused isolated_db fixture
* Remove unused transaction rollback functionality in db_session test fixture
* Rename /user REST resource to /users

Other changes
* Organize Makefile commands:
   * move setup-local to top
   * group openapi-spec with other CLI commands)
* Install pytest-lazy-fixture dev dependency
* Move empty_schema fixture to test_migrations, the only place it's used

## Context for reviewers
This PR consolidates the way we run the API server and the way we run background scripts to both use Flask's create_app factory. This removes the need to have two separate main() entrypoints, two separate ways to initialize the database, logger, etc. To do this, we leverage [Flask's built-in method of creating custom CLI commands](https://flask.palletsprojects.com/en/2.2.x/cli/#custom-commands) which itself is based on the [click library](https://click.palletsprojects.com/en/8.1.x/arguments/), a popular library for building Python command line applications that's more powerful than `argparse`. This allows us to remove the script_utils file.

This PR also renames the `api.route` package to `api.api` since we want to add "commands" to the user blueprint, so "route" is no longer an appropriate name that applies to everything in the blueprint. We could also consider other names like `api.components` or `api.blueprints`, and/or we could consider renaming the top level source code folder to something like `src` rather than `api`.

Finally, we made test_create_user_csv test things at using the test_cli_runner to be closer to the way we test routes using the test_client. This approach has more end-to-end test coverage.

Other tangential changes include making REST resources use plural names to follow REST resource naming conventions.

## Testing
Running `make cmd args='--help' prints out
```
... a bunch of general help stuff ...
Commands:
  routes  Show the routes for the app.
  run     Run a development server.
  shell   Run a shell in the app context.
  spec    Show the OpenAPI spec.
  user    User commands
```

Running `make cmd-user-create-csv args='--help'` prints out
```
Usage: flask user create-csv [OPTIONS]

  Create a CSV of all users and their roles

Options:
  --dir TEXT       Directory to save output file in. Can be an S3 path (e.g.
                   's3://bucketname/folder/') Defaults to current directory
                   ('.').
  --filename TEXT  Filename to save output file as. Defaults to
                   '[timestamp]-user-roles.csv.
  --help           Show this message and exit.
```

Actually running it produces a file like
```
"User Name","Roles","Is User Active?"
"Jon Doe","","True"
"Jane Doe","","True"
"Jack Dougherty","","True"
"User One","USER","True"
"Player One","USER","True"
"Player Two","USER","True"
"Admin One","ADMIN","True"
"Admin Two","ADMIN","True"
"Godzilla Dinosaur","ADMIN USER","True"
```
(these were the users i had in my db)


app/Makefile
```diff
@@ -33,6 +33,19 @@ else
 PY_RUN_CMD := docker-compose run $(DOCKER_EXEC_ARGS) --rm $(APP_NAME) poetry run
 endif
 
+FLASK_CMD := $(PY_RUN_CMD) flask --env-file local.env
```
chouinar, line 36: How would we run commands non-locally?

app/api/api/healthcheck.py
```diff
@@ -7,8 +7,8 @@
 from werkzeug.exceptions import ServiceUnavailable
 
 import api.adapters.db.flask_db as flask_db
-from api.route import response
-from api.route.schemas import request_schema
+from api.api import response
```
chouinar, line 10: Should we consider renaming the api folder to something else (src?) - I don't think api is accurate either as the CLI commands aren't an API.

app/openapi.yml
```diff
@@ -46,7 +46,7 @@ paths:
       tags:
       - Health
       summary: Health
-  /v1/user:
+  /v1/users:
```
chouinar, line 49: What is the purpose of the openapi file generated? Is it necessary, or just for a developer to see?

app/tests/api/scripts/test_create_user_csv.py
```diff
@@ -1,79 +1,90 @@
...
+from api.db import models
+from api.db.models.user_models import User
 from tests.api.db.models.factories import UserFactory
+from tests.lib import db_testing
 
 
 @pytest.fixture
-def tmp_file_path(tmp_path):
-    return tmp_path / "example_file.csv"
+def isolated_db(monkeypatch) -> db.DBClient:
```
chouinar: A few questions:
Why is this necessary? What is special about running CLI commands that requires this?

Why move it out of conftest? Presumably if someone using this repo wants to add more CLI commands and needed this, they'd move it back there so they don't copy-paste this fixture to each separate file - if it's only for script tests, could make a new conftest file in the tests/api/scripts folder (I think that makes it only available for subfolders/modules?)

app/Makefile
```diff
@@ -171,11 +176,21 @@ lint-security: # https://bandit.readthedocs.io/en/latest/index.html
...
+	$(FLASK_CMD) $(args)
+
+cmd-user-create-csv: ## Create a CSV of the useres in the database (Run `make cli-user-create-csv args="--help"` to see the command's options)
+	$(FLASK_CMD) user create-csv $(args)
+
+# Set init-db as pre-requisite since there seems to be a race condition
+# where the DB can't yet receive connections if it's starting from a 
+# clean state (e.g. after make stop, make clean-volumes, make openapi-spec)
+openapi-spec: init-db ## Generate OpenAPI spec
+	$(FLASK_CMD) spec --format yaml --output ./openapi.yml
```
chouinar, line 189: I had to deal with that race condition on NJ as the Oracle DB takes 60+ seconds to startup: https://github.com/navapbc/archive-nj-dol-ui-claimant-api/blob/main/wait-for-local-oracle.sh - it just checks if a particular log message appears every 5 seconds.

app/Makefile
lorenyu: ECS could run it directly with via `poetry run flask --app api.app <cmd>`.

Open to suggestions though. The challenge I ran into with the Makefile is that while docker-compose.yml has `env_file: local.env` which automatically loads the env file, running it on local host machine with `PY_RUN_APPROACH=local` won't automatically load the env vars.

I would love to move away from the `load_local_env_vars` function in our code. I think it's not very elegant to have `if environment == <x>` checks in the application code itself. I was actually a bit confused to see that load_local_env_vars was being called in production, and only realized upon digging into the function that it is a no-op based on the ENVIRONMENT env var.

app/api/api/healthcheck.py
lorenyu: Yeah I'm cool with `src`. At first I considered naming it `app`, but then it is confusing and breaks some files since in some files `app` is used for the local variable `Flask` app. I can go ahead and rename it to `src` unless you have any objections or other names you prefer.

app/openapi.yml
lorenyu: Good question I actually never thought about it. I just assumed the openapi file is some sort of spec that documents/captures the API, and can also be something that is used for other tools, like other tools can run checks on it or something. But I'm not really sure. Are you suggesting we get rid of it and just rely on the live /docs endpoint?

app/tests/api/scripts/test_create_user_csv.py
lorenyu: There's nothing special about CLI commands in general. But the create user csv command will dump all the users in the database into a CSV file, so any users that might have been added to the database in other tests will also show up in the file.

Previously we got around this by having tests include a [test_db_session](https://github.com/navapbc/template-application-flask/blob/5f07d0365220033d9dca7740b4425fd993f3dc1c/app/tests/conftest.py#L103-L122) fixture which started a nested transaction and automatically rolled back at the end of each test, but that only works for tests where we have some direct control over the session directly, but other tests like the ones in test_user_route use the flask app.test_client to make requests, which under the hood create a new session. If I wanted that new session under the hood to instead use an existing test session, that would require quite a bit of dependency injection that would also reduce my confidence that the test is running the same code that the prod code will. All that's to say, it's harder to automatically roll back db changes in tests that call the API in a somewhat end-to-end fashion, so if we wanted to have a test that has a fully isolated database schema, I think we'd explicitly create an "isolated_db" in that situation.

I don't feel too strongly about putting it in this file or leaving it in conftest. I guess I wasn't sure how common of a case it would be to have scripts that need to be fully isolated.

Thoughts?

app/Makefile
lorenyu: Oh interesting! I wonder if there's a better way to do that here as well.

app/Makefile
chouinar: I'm fine with moving the env var loading somewhere else.

I think the issue I had here was that we can't use `FLASK_CMD` anywhere non-locally as the env-file is in the command itself. Would it make more sense to do some sort of `if` statement in the makefile to add that bit to the command if you're running locally? That way when running non-locally it doesn't get used.

app/api/api/healthcheck.py
chouinar: `src` works. `app` would lead to `app.app` in the path which would be its own issue (might be worth renaming that top-level folder)

app/openapi.yml
chouinar: If we wanted it for reference that might be fine, but how do you make sure it stays up to date with schema changes in the code?

app/Makefile
lorenyu: I see that makes sense. How do I detect if Makefile is running locally? I'm caught up in a bit of chicken-egg. The `ENVIRONMENT=local` setting is defined in `local.env`. Should we make that specific environment variable something that is defined outside of `local.env`, either with `direnv` or something?

app/openapi.yml
lorenyu: There's an `Update OpenAPI spec` GitHub action that automatically updates the openapi.yml file if any of the code schemas change. That workflow was merged in this PR: https://github.com/navapbc/template-application-flask/pull/126

app/api/api/healthcheck.py
lorenyu: Ok renamed to `src`. Big change, hope i didn't break anything 🤞

app/tests/api/scripts/test_create_user_csv.py
chouinar: I think there's a separate issue that is at play that needs to be solved first. Each test needs to be independent from each other, otherwise they could conflict when running tests - like you're seeing. If I'm following right, and tinkering with the code right, the rollback doesn't happen between tests, meaning every test is done in a single transaction, rather than independent ones like it did before.

For example, I just pulled up one of the route tests, and the DB objects created from the prior tests are still present. That's going to cause issues (it's caused issues in the past for tests written incorrectly) and means running tests individually and all at once could behave differently.

Before the DB refactor, I believe this was handled by giving the DB session to the app itself, but now the session is fetched from a DB client it generates. There might be a way to get it working that way without any significant dependency injection or refactor:
- Make `create_app` take in an optional db_client (creating it if not supplied)
- The `app` test fixture takes in this `db_client` already, just pass it through
- Create a subclass of `db_client` which overrides the `get_session` method to return the `test_db_session` -might be a few ways to do this, but something to make that session return, perhaps moving the `test_db_session` logic into that method?

lorenyu: > Each test needs to be independent from each other, otherwise they could conflict when running tests - like you're seeing.

I agree that tests should be independent, but I think the responsibility should generally be on the test design. In other words, tests generally shouldn't care about what already exists in the database. If a test fails because there was already something in the database from a previous test, I think that's poor test design. I think the case of a service that dumps the entire database to a file is the exception, not the rule.

I actually think the current setup is more isolated than it was previously. Previously, all tests were running under the same transaction, and each test was using a separate nested transaction. I think that unnecessarily prevents the ability of testing code that creates new sessions and new transactions. I think the previous mechanism of having all tests run in a single transaction with nested transactions that automatically rollback is too much "magic", and I prefer a setup that's simpler to understand and forces developers to design their services to be more testable and to adopt simpler/more robust testing practices.

For example, the test_user_route.py tests are still pretty easy to understand and isolated from each other – a test can create a user via POST /users, then fetch/update the user using the uuid of the user. It shouldn't matter if there were existing users in the database or not.

chouinar: (irregardless of the below, does the DB rollback even do anything anymore if the schema gets dropped as you have it setup?)

I agree it comes down to test design, but I've seen several engineers get tripped up by this. Does it make more sense to setup tests in a way that reduces the underlying complexity or reduces the likelihood an engineer makes a test that only works when ran in certain ways?

I'd also really like to emphasize the performance issues that will happen with the isolated_db approach. Mass had that and explicitly moved most tests away from it because it gets incredibly slow to migrate the entire schema for 2k+ tests. I did the switch for most of the ~1k payments tests and brought the runtime down by about 10 minutes.

Is there some other way to do the rollback that is tied to the DB client rather than the nested session?

lorenyu: > (irregardless of the below, does the DB rollback even do anything anymore if the schema gets dropped as you have it setup?)

The `test_db_session` (the one that does the rollback) fixture is currently only used by the test_factories tests. I hadn't really gotten around to looking at the test_factories tests to see if we really need to rollback after each test, or if we can redesign the tests to be more agnostic of what's already in the db. If we remove the dependency on `test_db_session` from `test_factories.py`, then we can get rid of `test_db_session` completely. Currently, the fixture does still properly rollback after each test.

> I'd also really like to emphasize the performance issues that will happen with the isolated_db approach. Mass had that and explicitly moved most tests away from it because it gets incredibly slow to migrate the entire schema for 2k+ tests. I did the switch for most of the ~1k payments tests and brought the runtime down by about 10 minutes.

I totally agree. That's why 99% of the tests use the shared schema. The `db_client` fixture in conftest has scope="session" so it creates a single isolated_db for the entire test session.

The `isolated_db` fixture is only used for test_create_user_csv which has the behavior of dumping the entire database contents to a file, which is why that one requires an isolated db. That's the only test that requires an isolated db, everything else uses a shared one and therefore runs pretty fast.

I would actually love to get rid of the `isolated_db` fixture altogether. I think most projects won't even need it. How about we redesign the example backend script create-user-csv to do something more realistic (e.g. dump all users created after a certain time)? If we do that, then we don't need an isolated db to test it.

lorenyu: On second thought, maybe it's not unreasonable to assume that some small subset of tests would need an isolated_db. Maybe we could just call it out in docstring/comment that isolated_db is expensive and generally should only be used in situations where you absolutely need an isolated db.

lorenyu: It looks like the test_factories.py didn't even really rely on the rollback feature (the tests are already designed in an isolated manner), so I just removed the rollback functionality from `test_db_session` in the last commit since it's not really even being used.

chouinar: I think that sitting down and talking through this would be the best way forward. I consider it a requirement to isolate tests from each other (whether that's rollback or some other mechanism), I know this isn't explicitly due to this PR, but is something I would have pushed back on had I realized it was changed in one of the earlier PRs for the DB session.

lorenyu: I think the question is more about what level of isolation rather than whether to be isolated. The previous approach was isolated in some ways and actually less isolated in other ways.

I scheduled a meeting on monday to discuss live, and we can jot down our conversation notes into an ADR for others to see and/or chime in as well.

chouinar: Yes, I think it might be good to go through the types of isolation and what is worth it and what isn't. What types of isolation are worth discussing?
- isolating the DB schema creation has performance issues, so shouldn't be done for the average test
- it's not possible to isolate a DB sequence as they occur outside transactions - you could if made the tests reset sequences between tests, but that's probably more hassle than it's worth
- The DB data itself - isolating helps avoid collisions, but does require the test-specific rollback
- Non-db "static" objects should be either re-generated or otherwise immutable (eg. a base request that other tests override) to avoid one test changing it and that updated version being used
- what else?

lorenyu: yeah those sound right.

also, the isolation topic is somewhat closely tied with the topic of levels of abstraction i.e. what abstraction do we want to hide behind the Flask app / test framework, and what we want to expose to underlying application code.

For example, with the previous structure where we created transaction level isolation by starting a transaction and a nested transaction before running the test, it means that the application code can't have it's own concept of transactions. In particular, if the application code tried to do `session.begin()` to start a transaction, it will error with `sqlalchemy.exc.InvalidRequestError: A transaction is already begun on this Session.` If we want to make that work, we'll have to abstract away the concept of transactions from the service methods, and have the Flask application framework handle the creation of the transaction, which means we'll have to force every Flask request to be in exactly one transaction rather than let the service code decide when to start the transaction. I think that's a valid approach, but I worry that the boundaries of any particular "transaction" can be really specific to business logic, so it seems that the service code, rather than the framework, should be the one responsible for deciding when a transaction starts/ends, or whether a transaction is even needed at all.

chouinar: Yeah, that transaction problem was an issue for some of the payment tests on Mass.

What about truncating the tables after a test runs? I think something like:
```
    for table in Base.metadata.tables.values():
        table.delete()
```
Testing it with the NJ repo + removing the transaction pieces, it seems to work. Not sure if that would have any other consequences or considerations we'd want. Would also be worth check if there's a "better" way to do this rather than dozens of individual delete statements.

lorenyu: Truncating seems reasonable. It would still add a lot of unnecessary network requests for most tests that are already designed to not care about what was already in the database. Going along that same vein, what if we took the truncation idea and just did it on demand if a particular test needs a clean db? In other words, instead of my "isolated_db" fixture, the create-user-csv test would just truncate the User table at the beginning of the test.

app/api/api/healthcheck.py
lorenyu: Reverted the change and will do it in a subsequent PR. Created ticket #144

lorenyu:
@chouinar ok, I made the changes we discussed and created follow-on tickets for:

- #144 rename source code directory from api to src
- #145 create a document for testing best practices
- #143  add service level tests

app/Makefile
chouinar: Something in the setup-local command?

app/tests/api/scripts/test_create_user_csv.py
```diff
@@ -1,79 +1,58 @@
...
-
-def test_create_user_csv_local(isolated_db_factories_session, tmp_path, tmp_file_path):
-    # Same as above test, but verifying the file logic
-    # works locally in addition to S3.
-    db_records = [
-        UserFactory.create(first_name="A"),
-        UserFactory.create(first_name="B"),
+def prepopulate_user_table(enable_factory_create, db_session: db.Session) -> list[User]:
+    # First make sure the table is empty
+    db_session.query(User).delete()
```
chouinar: Maybe call out that they're from other tests?

chouinar (approved):
LGTM!

app/Makefile
lorenyu: Hmm yeah maybe. Anyways I created a ticket to revisit env vars more thoroughly https://github.com/navapbc/template-application-flask/issues/136

app/tests/api/scripts/test_create_user_csv.py
lorenyu: done!

PR #140, opened 2023-02-14
Add test for response.time

lorenyu:
## Ticket

n/a

## Changes
see title

## Context for reviewers
PR #134 added response.time to the response logs. i realize it might be nice to add a quick unit test for it. the unit test leverages sleep so it's not great but it's only a 100ms test and i figure we won't have many tests like this so 🤷 

## Testing
CI


chouinar (approved):
LGTM!

Issue #82, opened 2023-01-17
Upgrade to SQLAlchemy 2.0 once it moves out of beta


lorenyu:

Nava-JoshLong:
Using the version in main, the app builds fine, upgrading past it to 2.X causes errors. One of them is the following

```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/app/src/db/migrations/run.py", line 23, in up
    command.upgrade(alembic_cfg, revision)
  File "/home/joshlong/.cache/pypoetry/virtualenvs/template-application-flask-9TtSrW0h-py3.11/lib/python3.11/site-packages/alembic/command.py", line 322, in upgrade
    script.run_env()
  File "/home/joshlong/.cache/pypoetry/virtualenvs/template-application-flask-9TtSrW0h-py3.11/lib/python3.11/site-packages/alembic/script/base.py", line 569, in run_env
    util.load_python_file(self.dir, "env.py")
  File "/home/joshlong/.cache/pypoetry/virtualenvs/template-application-flask-9TtSrW0h-py3.11/lib/python3.11/site-packages/alembic/util/pyfiles.py", line 94, in load_python_file
    module = load_module_py(module_id, path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/joshlong/.cache/pypoetry/virtualenvs/template-application-flask-9TtSrW0h-py3.11/lib/python3.11/site-packages/alembic/util/pyfiles.py", line 110, in load_module_py
    spec.loader.exec_module(module)  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/app/src/db/migrations/env.py", line 20, in <module>
    from src.db.models import metadata  # noqa: E402 isort:skip
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/src/db/models/__init__.py", line 3, in <module>
    from . import base, user_models
  File "/app/src/db/models/user_models.py", line 21, in <module>
    class User(Base, IdMixin, TimestampMixin):
[...]

PR #141, opened 2023-02-15
Dockerfile staging and speed improvements

### LINKED ISSUE #23: Productionize Docker image
# Context

In order to productionize the [current Dockerfile](https://github.com/navapbc/template-application-flask/blob/main/app/Dockerfile), we need to:
* use non-root user (see [slack thread](https://nava.slack.com/archives/C03G1SWD9H7/p1664483403756849))
  * we don't need to make this an argument, we could probably hardcode it
* reduce the size of the built image by only copying over the necessary files (see [how PFML does it](https://github.com/navapbc/archive-massgov-pfml/blob/main/api/Dockerfile#L81-L120))

# References

* [PFML Dockerfile](https://github.com/navapbc/archive-massgov-pfml/blob/main/api/Dockerfile)

### LINKED ISSUE #37: Migrate poetry dev dependencies to new group syntax

### LINKED ISSUE #137: Pin Poetry versions to speed up build time
In our repo, the build times for the release image was taking ~10 minutes, with the poetry installs taking over 4 minutes. Upon researching, I found [this](https://github.com/python-poetry/poetry/issues/2094) thread where it says that when a poetry version is specified as a range of versions (ie `alembic = "^1.8.1"`), then poetry has to resolve any dependencies or issues for a range of versions. If you change it to a specified version (`alembic = "1.9.1"`), it can help speed up build times. Once we changed to pinned versions, it dropped the build time to ~6 minutes, with the poetry installs taking ~1 minute now. 

I have the versions pinned in local changes, but since this is a basic app, it only saves ~10 seconds. I'll be adding some docs explaining the tweaks and reasoning behind it so that other repos can make the changes and report back the time saved

Nava-JoshLong:
## Ticket

Resolves #137 
Resolves #37 
Resolves #23 
Vulnerability Scanner Configuration Files - From changes in the infra repo

## Changes
There are three main types of changes in this PR

### Vulnerability Scanner Configuration Files
- In the infra repo, we added a vulnerability scanner workflow that is called for PRs and merges to main to ensure that release images are as secure as we can make them
- We cannot use the ECR image scan in all cases, like when an image is based off of the `scratch` image
- The ECR image scan occurs after the image is pushed, so unless the image is removed if there are findings, it could still be used in production
- Some findings from the scanners need to be recasted as a lower level, or ignored because of one reason or another
- Each file specifies what is being changed, and why

### Tweaked Makefile
- Updated `app/Makefile` to have the required command and logic to build the release image

### Multi-stage Dockerfile
- The original `Dockerfile` had the local and release image as the same image, which included the testing packages which aren't needed for the release image
- The commit adds multi-stage builds to better separate out the different images and their configurations
- `base` is the base stage that the other images are based off of. This has the basic configuration for the app
- `dev` is the local version of the image stage. This image is used for testing and linting before pushing the image to ECR and uses the poetry groups `app` and `dev`
- `release-build` is the base for the release image. This image does not use the `dev` poetry group (the test suite packages), then builds the poetry wheel to use in the release image
- `release` is the image used in ECS, it is based on `scratch` to reduce the image size as well as any areas that could be used for vulnerabilities. The necessary files are explicitly copied over from `release-build`, so any new files will have to be copied over using the same method
- Updated the `docker-compose` file so that it knows to build the `dev` image locally instead of everything

### Poetry Changes
- Added comment to explain how to copy over non-python files (ie with a code-first swagger spec, you'd need to add `openapi.yml`)
- Poetry's latest versions uses groups to differentiate the required packages for different applications or images. There was a change to have the groups for this app
- The groups are split into three groups
- - Base -> These packages are the base for all groups, it is the base required packages required are for the application only, and will be in the release image
- - Dev -> These packages are for local only images, they include the linting and testing packages
- In the NJ project, we noticed that the poetry build times were slowing down the entire process. 
- - When we looked into it, we found out that when a version range is specified in poetry, poetry will need to resolve any issues or dependencies for all the package ranges
- - If we reduced the version ranges, or pinned to a specific version like in this PR, it would reduce the amount of the resolutions that poetry would need to perform
- - By pinning versions, we were able to the overall build time by ~4 minutes
- - Testing in the example app was able to reduce by 10 seconds, but other applications should see a greater reduction in time

## Context for reviewers
Testing the `dev` version of the image uses the same commands as before, `make init`. To test the release image, run the following command `make release-build; sleep 15; docker run --env-file ./app/local.env --name template -p 8081:8080 template-application-flask-app:latest`. This will fail since the app doesn't have a database to hit

To test the vulnerability configs locally, you will need to copy the file from the infra repo into the workflows directory. Then, you will need to install `act` and run it with `act push` which will simulate a push event in GitHub. The jobs run at the same time, so the output will step on each other, but you should see successes at the end of each job. Note that it might run into API limits, so you might need to run each job by using `act -j {JOB_NAME}` changing `{JOB_NAME}` for the name of the job in the workflow

## Testing
The image scan writes over it's output since each job is running at once, so it isn't clean to paste the output here. As for testing the release image, we would need to create a new env file that has the connection info to a local DB since we cannot use localhost as the container does not have the database built in


Makefile
```diff
@@ -0,0 +1,49 @@
+# PROJECT_NAME defaults to name of the current directory.
```
lorenyu: i'd rather not include the root Makefile here since the root Makefile is provided by template-infra, so it's going to be a maintenance burden to have to maintain both and make sure they are in sync

app/Dockerfile
```diff
@@ -2,26 +2,47 @@
...
+# Keep container packages up-to-date.
 RUN apt-get update \
     && apt-get install --no-install-recommends --yes \
     build-essential \
     libpq-dev \
     postgresql
 
+# Install poetry, the package manager.
+# https://python-poetry.org
+RUN pip install --no-cache-dir poetry==1.3.1
```
lorenyu, line 20: while we are here, can you add a comment explaining what `--no-cache-dir` does? i think the template/platform is a good opportunity to not only to spread best practices but also a way to educate project developers about the rationale behind those best practices

also are we sure we want to pin the poetry version? it seems like we should be ok with minor and patch upgrades of poetry i.e. `poetry~=1.3.1`

app/Dockerfile
```diff
@@ -2,26 +2,47 @@
...
+
+# setup user stuff
+ARG RUN_UID
+ARG RUN_USER
+
+RUN : "${RUN_USER:?RUN_USER and RUN_UID need to be set and non-empty.}" && \
+    [ "${RUN_USER}" = "root" ] || \
+    (useradd -mU --home "/home/${RUN_USER}" --uid ${RUN_UID} "${RUN_USER}" \
+    && mkdir /app \
+    && chown -R ${RUN_UID} "/home/${RUN_USER}" /app)
```
lorenyu: Can we add some comments explaining what this does?

app/Dockerfile
```diff
@@ -2,26 +2,47 @@
...
+USER ${RUN_USER}
+
+#-------------------------------------------------------------------------------
+# Development build environment
+#-------------------------------------------------------------------------------
+FROM build as dev
+
+ARG RUN_USER
+
+USER ${RUN_USER}
```
lorenyu: For my own understanding, why do we need `ARG RUN_USER` and `USER ${RUN_USER}` in every stage of the Dockerfile?

app/pyproject.toml
```diff
@@ -4,49 +4,66 @@ version = "0.1.0"
...
-isort = "^5.10.1"
-mypy = "^0.971"
-moto = {extras = ["s3"], version = "^4.0.2"}
-types-pytz = "^2022.2.1"
-coverage = "^6.4.4"
-Faker = "^14.2.0"
-factory-boy = "^3.2.1"
-bandit = "^1.7.4"
 
+[tool.poetry.group.app]
```
lorenyu: What's the reason for an `app` dependency group? It seems simpler to just have all prod dependencies be groupless, and only call out `dev` dependencies as a separate group

app/Dockerfile
```diff
@@ -2,26 +2,47 @@
...
+    (useradd -mU --home "/home/${RUN_USER}" --uid ${RUN_UID} "${RUN_USER}" \
+    && mkdir /app \
+    && chown -R ${RUN_UID} "/home/${RUN_USER}" /app)
+
+USER ${RUN_USER}
+
+#-------------------------------------------------------------------------------
+# Development build environment
+#-------------------------------------------------------------------------------
+FROM build as dev
```
lorenyu: More of a thought exercise, but what if we built `dev` FROM `app-build` rather than from `build`? It seems like `dev` is kind of similar to `app-build` but with some extra stuff (the `dev` dependencies in particular).

app/Dockerfile
```diff
@@ -33,5 +54,64 @@ ENV HOST=0.0.0.0
...
+RUN poetry install --no-root --with app --without dev
+
+COPY . ./
+
+# Install application dependencies
+# The name of the wheel file comes from pyproject.toml
+RUN poetry build --format wheel && poetry run pip install 'dist/template_application_flask-0.1.0-py3-none-any.whl'
+
+#-------------------------------------------------------------------------------
+# Application only
```
lorenyu: nit: the comment is a bit of a misnomer since the entire Dockerfile is for the application. Could we say something like "production image" or something

app/pyproject.toml
```diff
@@ -4,49 +4,66 @@ version = "0.1.0"
...
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.scripts]
+run-api = "api.__main__:main"
 db-migrate-up = "api.db.migrations.run:up"
 db-migrate-down = "api.db.migrations.run:down"
 db-migrate-down-all = "api.db.migrations.run:downall"
-create-user-csv = "api.scripts.create_user_csv:main"
+create-user-csv = "api.services.users.create_user_csv:main"
```
lorenyu: just sharing for awareness that i have a PR that intends to change this, so it's probably ok to leave this kind of broken for now https://github.com/navapbc/template-application-flask/pull/138

lorenyu (commented):
Leaving an initial review for now, since we probably need to figure out what to do with the root level Makefile. One of the challenges of splitting up the templates across separate repos.

Makefile
Nava-JoshLong: Makes sense, will be removed in next commit

app/Dockerfile
Nava-JoshLong: Added a blurb about it, but I'll throw it here too
```
# Install poetry, the package manager - https://python-poetry.org
# We use --no-cache-dir here because the pip caches normally, and if 
#  we don't cache, it will reduce the overall image size
# The version is pinned so that there is a known good configuration 
#  that we have better control over vs a range that can be used
```

app/Dockerfile
Nava-JoshLong: Since each stage is its own image, we need to tell it what user to run as in that stage. The default user is root otherwise. Adding this as a comment as well

app/pyproject.toml
Nava-JoshLong: In the NJ repo, we have the three groups so that we can run a python script to build the ECS config JSON that doesn't have the app or dev packages. We decided to do this since that script doesn't need anything special, but doesn't need the extra overhead of the other packages. I guess a better thing would to have a fourth group for something like this, called `scripts`, and add that stage to Dockerfile + the compose file?

app/Dockerfile
Nava-JoshLong: We could, but there isn't a ton of benefit that I can think of. The `dev` image is only built for local and linting+testing on pushes, so having another stage between `build` and `dev` wouldn't help a ton, and might add some wrinkles in needing more than just calling the `poetry install` command with both groups

app/Dockerfile
Nava-JoshLong: Changed in next commit

app/pyproject.toml
Nava-JoshLong: Gotcha, reverted in next commit

app/pyproject.toml
lorenyu: hm, that need seems awfully specific to that project so let's not carry that pattern/idea over here. let's default to the simplest thing that most projects will need which is just default prod dependencies + a dev group. i don't see a need to dynamically generate the ECS config JSON.

app/Dockerfile
lorenyu: got it, thanks for humoring me

.grype.yml
```diff
@@ -0,0 +1,26 @@
+# List of vulnerabilities to ignore for the anchore scan
```
lorenyu, line 1: this file looks to be the same as the one on template-infra so i think we can remove it, since it'll be copied over from template-infra when the project installs template-infra

.dockleconfig
```diff
@@ -0,0 +1,6 @@
+# This file is allows you to specify a list of files that is acceptable to Dockle
```
lorenyu, line 1: something that bugs me a bit, maybe we can discuss now or punt to solve in the future, is that the docker scanning is kind of split between template-infra and template-application-flask. the CI is in template-infra, but the configuration files are in the project. it's a little weird then because a project that uses template-application-flask but uses their own custom infra (doesn't use template-infra), will have this `.dockleconfig` file but they won't have any use for it.

maybe for now, can we add something in the README.md that explains that .dockeconfig and .hadolint.yaml are actually used to configure the vulnerability scanning defined in the template-infra repo, and link to the docs in that repo for more information. that way a developer looking at the flask template repo won't be confused and wondering what this file is all about.

.trivyignore
```diff
@@ -0,0 +1,9 @@
+# List of vulnerabilities to ignore for the trivy scan
```
lorenyu, line 1: i'm wondering if we can get rid of this file too since it looks to be the same as on template-infra

app/Dockerfile
```diff
@@ -2,29 +2,60 @@
...
+# The version is pinned so that there is a known good configuration 
+#  that we have better control over vs a range that can be used
+RUN pip install --no-cache-dir poetry==1.3.1
+
+# setup user stuff
+ARG RUN_UID
+ARG RUN_USER
+
+# The following logic creates the RUN_USER home directory and the directory where
+#  we will be storing the application in the image. This runs when the user is not root
```
lorenyu: i noticed in a few places you add an extra space in the comment e.g. `#  foo` instead of `# foo`, is that intentional?

app/Dockerfile
```diff
@@ -33,5 +64,76 @@ ENV HOST=0.0.0.0
...
+# Set the application working directory.
+WORKDIR /app
+
+# install dependencies
+# use a project-local virtualenv to be easy to find
+ENV POETRY_VIRTUALENVS_IN_PROJECT true
+COPY pyproject.toml poetry.lock ./
+RUN poetry install --no-root --with app --without dev
+
+COPY . ./
```
lorenyu: in this stage we do

```
WORKDIR /app
...
COPY . ./
```
but in dev we do
```
WORKDOR /app
COPY . /app
```
could we be consistent between the two (ether `COPY . .` or `COPY . /app`  since they do the same thing? i don't really have a preference between the two, but if we decide to do `COPY . /app` instead of `COPY . .` then let's also change the `COPY pyproject.toml poetry.lock ./` line to be `COPY pyproject.toml poetry.lock /app`)

app/Dockerfile
```diff
@@ -33,5 +64,76 @@ ENV HOST=0.0.0.0
...
+USER ${RUN_USER}
+
+# Set the application working directory.
+WORKDIR /app
+
+# install dependencies
+# use a project-local virtualenv to be easy to find
+ENV POETRY_VIRTUALENVS_IN_PROJECT true
+COPY pyproject.toml poetry.lock ./
+RUN poetry install --no-root --with app --without dev
```
lorenyu: Can we add a comment for why we're specifically copying `pyproject.toml` and `poetry.lock` even though right after this we copy the entire `.` directory with `COPY . .`. My understanding is that functionally it's redundant, but that this way we can separate out the dependencies as an initial container layer to improve the cache performance of builds since dependencies don't change as often as the source code.

app/Dockerfile
```diff
@@ -2,29 +2,60 @@
...
+USER ${RUN_USER}
+
 # Set the application working directory.
-WORKDIR /srv
+WORKDIR /app
 
 COPY pyproject.toml poetry.lock ./
 RUN poetry config virtualenvs.in-project false && poetry env use python
-RUN poetry install --no-root
+RUN poetry install --no-root --with app,dev
```
lorenyu: See other PR comment about adding a comment to explain why we first do `COPY pyproject.toml poetry.lock .` even though we do `COPY . .` soon after

app/Dockerfile
```diff
@@ -33,5 +64,76 @@ ENV HOST=0.0.0.0
...
+# If there is anything that needs to be ran as root, this is the spot
+
+# Change the user from root to ${RUN_USER}
+USER ${RUN_USER}
+
+COPY --from=app-build /etc/passwd /etc/group /etc/
+COPY --from=app-build /app/.venv /app/.venv
+
+# Include GPG for the DOR Import task.
+# Ideally this isn't included for every other ECS task/service, but we'll do it for now.
```
lorenyu: Please remove these lines.

app/pyproject.toml
Nava-JoshLong: Gotcha, removed in next commit

.grype.yml
Nava-JoshLong: I actually need to tweak the infra version since it has a python specific recast in it, the `certifi` line. Should I remove this file and leave it in infra, or keep it and tweak the infra version?

.dockleconfig
Nava-JoshLong: Makes sense, adding a blurb in the README in the next commit

app/Dockerfile
Nava-JoshLong: Yeah, I usually do that to show that it is a continuation of the previous comment. Not too sure where I picked it up, but I can remove that extra space if it is an issue

app/Dockerfile
Nava-JoshLong: I'll run the bottom version to see if there was a reason, and then make a commit with whichever version works in both cases, I forget why I did it like that

app/Dockerfile
Nava-JoshLong: Removed in next commit

.grype.yml
lorenyu: oh i see. so you're suggesting that we move the python specific thing from the infra repo to this repo? that makes sense.

app/Dockerfile
lorenyu: Oh I see. The designer in me thinks it looks off but I don't have strong feelings. Indenting the first line of a paragraph is more typical in traditional articles/papers rather than a hanging indent. Like:

```
#   Lorem ipsum
# dolores foo bar
# cat dog mouse.
#   Lorem ipsum
# dolores foo bar
# cat dog mouse.
```
 In code I think it's also visually easy to read justified text with an extra line between paragraphs.
```
# Lorem ipsum
# dolores foo bar
# cat dog mouse.
#
# Lorem ipsum
# dolores foo bar
# cat dog mouse.
```
slight preference for the latter for visual consistency but it's not something that really matters at the end of the day

.grype.yml
Nava-JoshLong: Yeah, as we continue to build in this repo, there will probably be other things we need to recast or safelist that are just for this repo, so I think we should keep the config files here since they'd fix those issues

.trivyignore
Nava-JoshLong: As we continue to build in this repo, there will probably be other things we need to recast or safelist that are just for this repo, so I think we should keep the config files here since they'd fix those issues

app/Dockerfile
Nava-JoshLong: Swapped to just using the `COPY . /app` since it covered everything

lorenyu: Would there a performance gain to having it separate though? My understanding is that with the docker build cache, if the dependencies in pyproject.toml and poetry.lock don't change, Docker could reuse that layer, and only need to rebuild the layer with the source files?

Nava-JoshLong: There might be a slight gain in having them separate, but I think it would be on the local version only since I'm not sure how much GitHub caches between builds. I know there is a Docker cache action for workflows, but I think it is still in beta right now

.hadolint.yaml
```diff
@@ -0,0 +1,10 @@
+# List of settings and ignore or safelist findings for the hadolint scanner
+
+# For more information on any settings you can specify, see the actions' documentation here
+# https://github.com/hadolint/hadolint#configure
+override:
+  info: 
+    # Casts the apt-get install <package>=<version> finding as info
+    # We have this set since there is no way to specify version for
+    #  build-essentials in the Dockerfile
+    - DL3008
```
lorenyu: super small nit could we add newlines to the end of the config files to avoid the annoying red circle in github
image

app/Dockerfile
```diff
@@ -33,5 +61,73 @@ ENV HOST=0.0.0.0
...
+COPY . /app
+RUN poetry install --no-root --without dev
+
+# Install application dependencies
+# The name of the wheel file comes from pyproject.toml
+RUN poetry build --format wheel && poetry run pip install 'dist/template_application_flask-0.1.0-py3-none-any.whl'
+
+#-------------------------------------------------------------------------------
+# Production Application Image only
+#-------------------------------------------------------------------------------
```
lorenyu: tiny nit: could we reduce the length of the comment section separator so it doesn't wrap in PR view?

my suggestion would be something like

```suggestion
######################
## Production Image ##
######################
```

app/Dockerfile
```diff
@@ -33,5 +61,73 @@ ENV HOST=0.0.0.0
...
+COPY --from=app-build /lib/* /lib/
+COPY --from=app-build /usr/local/lib/ /usr/local/lib/
+COPY --from=app-build /usr/local/include/python* /usr/local/include/
+COPY --from=app-build /lib64/* /lib64/
+COPY --from=app-build --chown=${RUN_USER} /tmp/ /tmp/
+
+ENV PATH="/app/.venv/bin:$PATH"
+
+# run the application
+CMD ["run-api"]
```
lorenyu, line 144: I don't really understand how this works, how does the image know to run the poetry script run-api without it being `CMD ["poetry", "run", "run-api"]`?

app/Dockerfile
```diff
@@ -33,5 +61,73 @@ ENV HOST=0.0.0.0
...
 # https://python-poetry.org/docs/basic-usage/#installing-dependencies
 RUN poetry install
+
 # Run the application.
 CMD ["poetry", "run", "python", "-m", "api"]
+
+#-------------------------------------------------------------------------------
+# Non-development build environment
+#-------------------------------------------------------------------------------
+FROM build as app-build
```
lorenyu: nit: could we name this `release-builder` since this is really just used as a base for the `release` stage?
```suggestion
#####################
## Release Builder ##
#####################
FROM build as release-builder
```

app/Dockerfile
```diff
@@ -2,29 +2,57 @@
...
+    [ "${RUN_USER}" = "root" ] || \
+    (useradd -mU --home "/home/${RUN_USER}" --uid ${RUN_UID} "${RUN_USER}" \
+    && mkdir /app \
+    && chown -R ${RUN_UID} "/home/${RUN_USER}" /app)
+
+USER ${RUN_USER}
+
+#-------------------------------------------------------------------------------
+# Development build environment
+#-------------------------------------------------------------------------------
```
lorenyu: nit shorten to
```suggestion
###############
## Dev Image ##
###############
```

app/Dockerfile
```diff
@@ -2,29 +2,57 @@
...
 # The build stage that will be used to deploy to the various environments
-# needs to be called `release` in order to integrate with the repo's
-# top-level Makefile
-FROM python:3-slim AS release
-
-# Install poetry, the package manager.
-# https://python-poetry.org
-RUN pip install poetry
+# This is what the other stages will call as their base
+FROM python:3.11-slim-bullseye AS build
```
lorenyu: nit: since this stage doesn't really build anything (it only installs same base tools like poetry and apt-get stuff), can we name this something else, like `build-tools` or `base` or something
```suggestion
FROM python:3.11-slim-bullseye AS build-tools
```

.hadolint.yaml
Nava-JoshLong: In latest commit

app/Dockerfile
```diff
@@ -2,29 +2,57 @@
...
+
+# Change the user from root to ${RUN_USER}
+USER ${RUN_USER}
+
 # Set the application working directory.
-WORKDIR /srv
+WORKDIR /app
 
-COPY pyproject.toml poetry.lock ./
+COPY . /app
```
lorenyu, line 53: If there was a performance gain for local docker builds to split up the "COPY dependencies" layer from the "COPY source files layer", should we keep that for the `dev` stage (and add a comment about it)? I think even if it saves a few seconds it could add up over the course of many local builds.

app/Makefile
```diff
@@ -33,15 +33,37 @@ else
 PY_RUN_CMD := docker-compose run $(DOCKER_EXEC_ARGS) --rm $(APP_NAME) poetry run
 endif
 
+# Docker user configuration
```
lorenyu, line 36: Could you say something about the motivation for this? I think Tanner had a good explanation [in slack](https://nava.slack.com/archives/C03G1SWD9H7/p1670961910976749?thread_ts=1670888895.529179&cid=C03G1SWD9H7)

app/Makefile
```diff
@@ -33,15 +33,37 @@ else
...
+export RUN_USER
+export RUN_UID
+
 ##################################################
 # API Build & Run
 ##################################################
 
 build:
-	docker-compose build
+	docker-compose build $(APP_NAME)
```
lorenyu, line 75: Doesn't this mean that `main-db` doesn't build?

app/Makefile
```diff
@@ -33,15 +33,37 @@ else
...
 
 start:
-	docker-compose up --detach
+	docker-compose up --detach $(APP_NAME)
```
lorenyu, line 78: Similar question, we need `main-db` to run too in order for the app to work, doesn't specifying $(APP_NAME) cause `main-db` not to run?

app/Dockerfile
```diff
@@ -33,5 +61,73 @@ ENV HOST=0.0.0.0
...
+# If there is anything that needs to be ran as root, this is the spot
+
+# Change the user from root to ${RUN_USER}
+USER ${RUN_USER}
+
+# Set the application working directory.
+WORKDIR /app
+
+# Install dependencies
+# Use a project-local virtualenv to be easy to find
```
lorenyu: Could we be more specific about where the virtualenv gets created and add a link to the poetry docs
```suggestion
# Install dependencies
# Set virtualenv location to be in project to be easy to find
# This will create a virtualenv in /app/.venv/
# See https://python-poetry.org/docs/configuration/#virtualenvsin-project
# See https://python-poetry.org/docs/configuration/#using-environment-variables
```

app/Dockerfile
```diff
@@ -33,5 +61,73 @@ ENV HOST=0.0.0.0
...
+ENV POETRY_VIRTUALENVS_IN_PROJECT true
```
lorenyu: the `ENV var val` alternative syntax is discouraged according to Docker docs and may be removed in the future, let's use the `ENV var=val` syntax (https://docs.docker.com/engine/reference/builder/#env)
```suggestion
ENV POETRY_VIRTUALENVS_IN_PROJECT=true
```

app/Dockerfile
```diff
@@ -33,5 +61,73 @@ ENV HOST=0.0.0.0
...
+COPY . /app
+RUN poetry install --no-root --without dev
+
+# Install application dependencies
+# The name of the wheel file comes from pyproject.toml
+RUN poetry build --format wheel && poetry run pip install 'dist/template_application_flask-0.1.0-py3-none-any.whl'
```
lorenyu: Is this the command that actually builds the `run-api` binary in `/app/.venv/bin`? I can't seem to figure it out for certain, but if I'm inferring correctly, having poetry build the python wheel is what actually creates the binaries that are defined in pyproject.toml. If my understanding is correct can we make the comment reflect that?
```suggestion
# Build the application binary (python wheel) defined in pyproject.toml
RUN poetry build --format wheel && poetry run pip install 'dist/template_application_flask-0.1.0-py3-none-any.whl'
```
I think the "# install application dependencies" comment belongs with the previous command

lorenyu (commented):
This is great thanks for being patient with me. I have a bunch of nits. Apologies for the nitpicks, but since most people using the template aren't going to be infra experts, I want to make sure the comments and stuff are as clear and precise (and educational) as possible.

app/Dockerfile
Nava-JoshLong: Changed in next commit, and added blurb for `app-build`

app/Dockerfile
Nava-JoshLong: I'll add a comment about it, but there are two things creates the directory for the app, and then copy it over. 
 - In `app-build`, we use a virtual environment and wheel to create the files in the working directory, `app`, named `.venv`. This directory contains the files needed for the app, and any non-python code that needs to be copied over, a la `openapi.yml`, would need to be explicitly listed in the `pyproject.toml` file in the `include` line at the top of the file
 - In `release`, the files in `app/.venv` is copied over from the line `COPY --from=app-build /app/.venv /app/.venv`. The python module (ie `requests`, 'logger`, etc) directories are required are copied over in a later command `COPY --from=app-build /usr/local/include/python* /usr/local/include/`

Once both of these are set, we tell the path that it can run from `app/.venv/bin` and it knows, from being built as a wheel, to run the scripts specified in `pyproject.toml` as actual commands like this

app/Dockerfile
Nava-JoshLong: In the next commit

app/Dockerfile
Nava-JoshLong: In the next commit

app/Dockerfile
Nava-JoshLong: I like `base` best, changing to that in the next commit

app/Makefile
Nava-JoshLong: Kind of? In the `build` command, we're just building the image, and we aren't configuring the DB with a Dockerfile, so it doesn't really affect it one way or another if we have the specification for the image or not. It somewhat saves time because Docker doesn't need to check for configs for the DB

app/Makefile
Nava-JoshLong: No, in the `docker-compose.yml` file, we tell `main-app` that it is dependent on `main-db`, so it will build the DB first. I had this here for in case there are different docker-compose files. You could tell it to start the specified app, instead of it starting all apps

app/Dockerfile
Nava-JoshLong: Added in next commit

app/Dockerfile
Nava-JoshLong: Good to know, in the next commit

app/Dockerfile
Nava-JoshLong: Yes, this is that command that builds in `app/.venv` and what has the binaries. And I like that, it is more clear, in the next commit

app/Dockerfile
Nava-JoshLong: There could be a slight performance gain since it would be cached, but we wouldn't be updating the `pyproject` or `poetry` files super often, since we should be using `dependabot` to update these files. I can split it back up, but IDK if it will save much time overall

app/Makefile
Nava-JoshLong: Do you mean change the comments in the code a bit to make it clearer + what Tanner said?

app/Dockerfile
lorenyu: I get it thanks!

btw do we actually need to include openapi.yml in the release build? I think the swagger docs just rely on the Marshmallow schemas and don't actually need the openapi.yml file. (the template uses apiflask which is a code-first approach to API rather than a spec-first approach)

app/Makefile
lorenyu: Oh I get it thanks

app/Makefile
lorenyu: Yeah exactly. The comment explains what it's doing (configure Docker users) but not why (we don't want Dockerfile to run as root user). I'm not sure the best place to put the comment, here or in the Dockerfile, but somewhere it would be useful to explain what Tanner described - i.e. the thing about mounted volumes needing to be owned by the same UID, and the thing about not using the root user in Dockerfile

app/Dockerfile
Nava-JoshLong: Good question, I am not 100% sure, but I'd think it might not be needed then? So, if I remove that file and run it locally, the code will generate the `/docs` endpoint?

lorenyu: I think so

Nava-JoshLong: Yeap, it worked. I commented out the code and added a blurb about when it is needed

app/Makefile
Nava-JoshLong: Slightly tweaked the comments to try and make it clearer

lorenyu:
btw i updated the PR description ticket section. if you put "Resolves #[ticket number]" it will automatically move the issue to the "Done" column in Github Projects

Nava-JoshLong:
> btw i updated the PR description ticket section. if you put "Resolves #[ticket number]" it will automatically move the issue to the "Done" column in Github Projects

Thanks! I wasn't sure how to do that, so that helps. I updated the description section a bit to reflect the changes since the PR was opened

app/Dockerfile
```diff
@@ -2,29 +2,57 @@
...
+
+# setup user stuff
+ARG RUN_UID
+ARG RUN_USER
+
+# The following logic creates the RUN_USER home directory and the directory where
+# we will be storing the application in the image. This runs when the user is not root
+RUN : "${RUN_USER:?RUN_USER and RUN_UID need to be set and non-empty.}" && \
+    [ "${RUN_USER}" = "root" ] || \
+    (useradd -mU --home "/home/${RUN_USER}" --uid ${RUN_UID} "${RUN_USER}" \
```
lorenyu: nit: can we use the long-form version of the flags? for automated code i prefer to optimize for readability, i think the short flags are useful when we're doing things interactively

```suggestion
    (useradd --create-home --create --user-group --home "/home/${RUN_USER}" --uid ${RUN_UID} "${RUN_USER}" \
```

Nava-JoshLong:
@lorenyu @chouinar Just a heads up, there are some things that will be broken in future module versions. I rolled back to the versions to something that works, but future versions are going to have issues. Keeping this as a running list

- The newer versions of `SQLAlchemy` will break [this line](https://github.com/navapbc/template-application-flask/blob/main/app/src/adapters/db/clients/postgres_client.py#L60). Looks like the module changes where `dbapi_connection` is located
- `alembic` has some issues with the `run.py` file in `db/migrations`, looks to be the function changed what it is expecting
- `APIFlask` isn't able to write the openapi file in the directory in the newer versions, not too sure why, it looks to be a permissions error, but I tried different levels of permissions in the directory and couldn't get it to write. Looks to still be an issue, even though I rolled back to the version in main?

Still kicking this around while I am juggling other things

lorenyu:
@Nava-JoshLong are you referring to future versions of SQLAlchemy or other packages? If you're referring to SQLAlchemy, could you add those notes to this ticket: #82?

lorenyu:
@Nava-JoshLong you still working on this?

Nava-JoshLong:
@lorenyu I am, but I had to put it on the back burner since I have other work that popped up. I ran into an issue where somehow the workflows have errors, and I can't reproduce them locally. I rolled some changes back locally, and still can't see where the issues are coming from. 

My plan is to try and split the overall changes up into multiple PRs in one of the NJ repos, and see if the same thing happens. If that works, I was going to close this PR, and do the same thing where I split it up and see if that works. The Freehold repo doesn't have these issues at all, so I'm not sure where or why these issues are happening. The only thing I can think of is that repo wasn't originally built off of this one, so there is some module that is causing all of my headaches

lorenyu:
sounds good thanks!

Nava-JoshLong:
Closing this to split into smaller future PRs to address errors we were seeing

PR #146, opened 2023-02-21
Rename application package to src from api

### LINKED ISSUE #144: Rename api to src
as part of PR #138 we renamed the route package to api, so to avoid name conflicts we should rename the top level source directory to src.

lorenyu:
## Ticket

Resolves #144 

## Changes
see title

## Context for reviewers
In PR #138 we renamed the route package to api, which makes it api.api. This PR renames the top level package to src to avoid naming issues and to represent the idea that the application contains other things like CLI commands / background jobs. We considered naming it app, but that would conflict with the app folder at the repo root, and it would also create ambiguity with the flask app variable.

## Testing
CI


chouinar (approved):
LGTM!

PR #152, opened 2023-03-07
Change DBClient setup to make multiple clients possible

chouinar:
## Ticket

https://github.com/navapbc/template-application-flask/issues/151

## Changes
Modified DBClient to instead take in a DBEngine class which handles setup and connection verification.

Modified the flask_db code to allow for specifying which DBClient to fetch a session from, but defaulting to just have one.

## Context for reviewers
On my project, we're soon going to need two DB clients, one for a Postgres DB, and one for an Oracle DB. This makes doing that possible.

Note that in doing this I noticed a few additional things we might want to consider, although I've left them out of this PR to keep it focused:
- Migrations are still tied to the postgres DB client
- Multiple clients of the same type (ie. two postgres) wouldn't be configurable, we'd need to adjust DBConfig and have that pass in

As there might be some discussion still, I haven't updated all the docs, especially anything regarding the engine. Will do that if this solution looks to be solid

## Testing
`make test` all tests pass - added and moved a few based on the changes


app/src/adapters/db/__init__.py
```diff
@@ -9,7 +9,7 @@
 Usage:
     import src.adapters.db as db
 
-    db_client = db.init()
+    db_client = db.init(PostgresEngine())
```
lorenyu: to make the example more complete could we add in the usage example where the PostgresEngine comes from e.g.

```
from src.adapters.db.engines import PostgresEngine
```

app/src/adapters/db/client.py
```diff
@@ -8,16 +8,11 @@
...
-import urllib.parse
-from typing import Any
 
-import psycopg2
 import sqlalchemy
-import sqlalchemy.pool as pool
 from sqlalchemy.orm import session
 
-from src.adapters.db.config import DbConfig, get_db_config
+from src.adapters.db.engine.db_engine import DbEngine
```
lorenyu: i realize we have `DbConfig` and `DBClient` so we have some inconsistent capitalization. since DB is an acronym i bias towards `DBEngine` instead of `DbEngine`. do you agree? for example, SQLAlchemy the library is capitalized as SQLAlchemy not SqlAlchemy

app/src/adapters/db/client.py
```diff
@@ -37,16 +32,17 @@ class DBClient:
     It has methods for getting a new connection or session object.
     """
 
+    _db_engine: DbEngine
     _engine: sqlalchemy.engine.Engine
```
lorenyu: if we're going to abstract the concept of the engine into a DBEngine class, then every module outside of db_engine.py should only interact with DBEngine and not with the underlying sqlalchemy.engine.Engine class, so I'd remove the second instance variable here.

app/src/adapters/db/client.py
```diff
@@ -37,16 +32,17 @@ class DBClient:
...
 
-    def __init__(self) -> None:
-        self.config = get_db_config()
-        self._engine = _create_db_engine(self.config)
+    def __init__(self, db_engine: DbEngine) -> None:
```
lorenyu: take a look at the `binds` parameter to the `Session` constructor: https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.params.binds

it looks like a single session can use multiple engines, where different ORM classes are mapped to the appropriate engine (e.g. all the models for Oracle tables can be mapped to the oracle engine and all the models for the Postgres tables can be mapped to the postgres engine). I wonder if that means that they support transactions and rollbacks across the two db systems, which if that's possible/true, might be pretty nice/powerful to ensure certain guarantees about data integrity (don't commit to one db if the other db didn't commit successfully).

if that's actually true then maybe we could reconsider having multiple db clients, and instead have multiple engines in a single db client. thoughts?

lorenyu: Related, this looks interesting:

> In the more common case, there are typically base or mixin classes that can be used to distinguish between operations that are destined for different database connections. The [Session.binds](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.params.binds) argument can accommodate any arbitrary Python class as a key, which will be used if it is found to be in the __mro__ (Python method resolution order) for a particular mapped class. Supposing two declarative bases are representing two different database connections:

```
from sqlalchemy.orm import DeclarativeBase
from sqlalchemy.orm import Session


class BaseA(DeclarativeBase):
    pass


class BaseB(DeclarativeBase):
    pass


class User(BaseA):
    ...


class Address(BaseA):
    ...


class GameInfo(BaseB):
    ...


class GameStats(BaseB):
    ...


Session = sessionmaker()

# all User/Address operations will be on engine 1, all
# Game operations will be on engine 2
Session.configure(binds={BaseA: engine1, BaseB: engine2})
```

from this doc https://docs.sqlalchemy.org/en/20/orm/persistence_techniques.html#session-partitioning

app/src/adapters/db/engine/db_engine.py
```diff
@@ -0,0 +1,18 @@
+import abc
+
+import sqlalchemy
+
+
+class DbEngine(abc.ABC, metaclass=abc.ABCMeta):
```
lorenyu: I mentioned something similar in a comment above, but makes sense to say it here too: if we're going to create a DBEngine abstraction, i'd like it to "replace" the sqlalchemy.engine.Engine abstraction so that outside of this module we only interact with DBEngine and not the underlying sqlalchemy.engine.Engine. In that case, we'll need to re-export / expose a subset of the sqlalchemy.engine.Engine methods that we use, and then define a few custom ones like "check_db_connection".

So I imagine this class would look like

```python
class DBEngine(abc.ABC):
  _engine: sqlalchemy.engine.Engine

  def __init__(self):
    self._engine = self._configure_engine()

  @abc.abstractmethod
  def configure_engine(self) -> sqlalchemy.engine.Engine:
    raise NotImplementedError()

  def connect(self):
    return self._engine.connect()

  @abc.abstractmethod
  def check_db_connection(self, conn: db.Connection) -> None:
    raise NotImplementedError()
```

then in client.py we'll need to change how we instantiate the `Session` to instead use a connection i.e. `get_session` should return `Session(bind=db_engine.connect(), ...)` rather than `Session(bind=db_engine._engine, ...)` so that we don't leak private variables to other modules.

app/src/adapters/db/engine/db_engine.py
```diff
@@ -0,0 +1,18 @@
...
+    @abc.abstractmethod
+    def build_engine(self) -> sqlalchemy.engine.Engine:
+        pass
+
+    @abc.abstractmethod
+    def check_db_connection(self, conn: sqlalchemy.engine.Connection) -> None:
```
lorenyu: nit, we currently define the type `Connection = sqlachemy.engine.Connection` in `client.py`, but now that we're extracting out the engine.py module, the Connection concept belongs more with the engine.py module rather than in the client.py module, so let's move that type definition to this file and simplify this type hint as

```suggestion
    def check_db_connection(self, conn: Connection) -> None:
```

lorenyu (commented):
Thanks for doing this so fast! Left some comments and also saw a new idea from the SQLAlchemy docs that might be worth a discussion

app/src/adapters/db/client.py
chouinar: Let's assume this works relatively easily, what is more intuitive to understand as a developer?

I'd argue having explicitly defined DB engines makes more sense to me. "I have session A that works with DB A, and session B that works with DB B" rather than a single session that works under the hood to do it.

I think anything to try and make this work requires it's own dedicated investigation, and we'd want multiple DBs setup to be able to test and validate the behavior, which we don't currently have (in NJ, this is prep work for eventually having that). I'm just trying to get the configuration setup to make multiple DBClients possible right now, I don't have the bandwidth to go beyond that right now.

lorenyu: yeah, i think our initial approach was easier to understand how it works, and until we see a more complex use case it's not worth overengineering it

app/src/adapters/db/engine/db_engine.py
chouinar: Hmm, if I do this, I end up with DBClient just being:

```
class DBClient:
    _db_engine: DbEngine

    def __init__(self, db_engine: DbEngine) -> None:
        self._db_engine = db_engine

    def get_connection(self) -> Connection:
        return self._db_engine.get_connection()

    def get_session(self) -> Session:
        return self._db_engine.get_session()

```

At that point, would it make more sense to not have client + engine, and just have a base client + derived clients (eg. PostgresDBClient)? The way this is setup, the DBClient doesn't actually do anything but call underlying methods

lorenyu: yeah i'd be fine collapsing DBClient and DBEngine into just DBClient, having DBClient be an abstract class with different concrete implementations. the distinction was pretty blurry to me to begin with.

lorenyu: the reason i originally suggested having engine be separate only made sense when it was just one function like the build_engine, in which case we might not have needed a whole class, just different ways to build different engines. but then we needed to add the check connection, and the connect, so the lines got blurry. sorry for pulling you in this direction in the first place, i think your original instinct ended up being the right one for our situation now.

chouinar: Yeah, I thought the separate engine made sense until we actually tried to do it. Nothing major to have to switch back, just copy-pasting a bunch of functions really. I switched it back in the latest commit

chouinar:
@lorenyu - let me know if this looks right, and I'll tidy up the comments and docs a bit more

app/src/adapters/db/flask_db.py
```diff
@@ -73,33 +80,42 @@ def get_db(app: Flask) -> DBClient:
         def health():
             db_client = flask_db.get_db(current_app)
     """
-    return app.extensions[_FLASK_EXTENSION_KEY]
+    return app.extensions[client_name]
```
lorenyu: just realized that i've been mixing up the "client_name" and the "flask_extension_key". The "client_name" needs to be unique in the context of db clients, but the flask extension key needs to be unique across all flask extensions, so i think we should add a prefix to namespace the key. maybe something like:

```
_FLASK_EXTENSION_KEY_PREFIX = "db-"
DEFAULT_CLIENT_NAME = "default"

def get_db(app, client_name: str = DEFAULT_CLIENT_NAME):
  flask_extension_key = f"{_FLASK_EXTENSION_KEY_PREFIX}{client_name}"
  return app.extensionions[flask_extension_key]
```

app/src/adapters/db/flask_db.py
```diff
@@ -73,33 +80,42 @@ def get_db(app: Flask) -> DBClient:
...
+        @with_db_session()
         def foo(db_session: db.Session):
             ...
 
-        @with_db_session
+        @with_db_session()
         def bar(db_session: db.Session, x, y):
             ...
+
+        @with_db_session(client_name="not_the_primary_client")
```
lorenyu: nit, could we have the secondary client example be based on some sort of realistic-sh scenario to give engineers some sort of intuition for when/how they might use it, e.g. something like `client_name="legacy_db"`

app/src/adapters/db/flask_db.py
```diff
@@ -73,33 +80,42 @@ def get_db(app: Flask) -> DBClient:
...
+        def fiz(db_session: db.Session, x, y, z):
+            ...
     """
 
-    @wraps(f)
-    def wrapper(*args: Any, **kwargs: Any) -> T:
-        with get_db(current_app).get_session() as session:
-            return f(session, *args, **kwargs)
+    def decorator_func(f: Callable[Concatenate[db.Session, P], T]) -> Callable[P, T]:
```
lorenyu: nit, i thought a decorator is always a function, so maybe just call it `decorator` instead of `decorator_func`
(btw don't feel too strongly about this comment)

app/src/adapters/db/postgres_client.py
```diff
@@ -0,0 +1,146 @@
+import logging
```
lorenyu: when you're cleaning up some comments would you mind adding one sentence blurb explaining what this file is i.e. it provides the a db client that interacts with a postgres database.

app/src/adapters/db/postgres_client.py
```diff
@@ -0,0 +1,146 @@
...
+logger = logging.getLogger(__name__)
+
+
+class PostgresDBClient(DBClient):
+    def __init__(self) -> None:
+        self.db_config = get_db_config()
+        super().__init__()
+
+        if self.db_config.check_connection_on_init:
+            self.check_db_connection()
```
lorenyu: nit: could we move references to db_config into `_configure_engine`? feels like it belongs there since configure engine has "configure" in the name, and it also feels clean for now to not need to save the db_config as an instance variable, and we also won't need to override the constructor anymore too.

app/src/adapters/db/__init__.py
```diff
@@ -9,7 +9,7 @@
 Usage:
     import src.adapters.db as db
 
-    db_client = db.init()
+    db_client = db.init_postgres_client()
```
lorenyu: now that we have separate clients i feel like we don't really need the `init` abstraction anymore, what do you think about just creating the client directly i.e.

```
from src.adapters.db.clients import PostgresDBClient

db_client = PostgresDBClient()
```

app/src/adapters/db/postgres_client.py
lorenyu: a small preference to put postgres_client in a folder `src/adapters/db/clients/postgres_client.py`, even though there's only one, just to indicate that that's where new clients should go if a project aims to define separate clients.

lorenyu (commented):
Looks great!

app/src/adapters/db/postgres_client.py
chouinar: I was setting the init up that way because I'm imagining a scenario where we have two clients/DBs that are of the same type.

To make that work, you'd need to make init take in the DB config, but otherwise would work the same. I could make that change, but was just getting it most of the way there at the moment. Thoughts?

chouinar: I adjusted the constructor now

lorenyu: that's a good point, it would be nice to override environment configs with a config parameter. happy to defer that to later then

app/src/adapters/db/clients/postgres_client.py
```diff
@@ -0,0 +1,149 @@
...
+logger = logging.getLogger(__name__)
+
+
+class PostgresDBClient(DBClient):
+    """
+    An implementation of a DBClient for connecting to a Postgres DB
+    as configured by parameters passed in from the db_config
+    """
+
+    def __init__(self, db_config: DBConfig | None = None) -> None:
```
lorenyu: this makes me think that `DBConfig` is specific to `PostgresDBClient` so maybe we should rename it to `PostgresDBConfig`? but happy to defer that convo for now

app/src/adapters/db/clients/postgres_client.py
```diff
@@ -0,0 +1,149 @@
...
+    def _configure_engine(self) -> sqlalchemy.engine.Engine:
+        # We want to be able to control the connection parameters for each
+        # connection because for IAM authentication with RDS, short-lived tokens are
+        # used as the password, and so we potentially need to generate a fresh token
+        # for each connection.
+        #
+        # For more details on building connection pools, see the docs:
+        # https://docs.sqlalchemy.org/en/13/core/pooling.html#constructing-a-pool
+        def get_conn() -> Any:
+            return psycopg2.connect(**get_connection_parameters(self.db_config))
```
lorenyu: at this point i think it's a bit more opaque to the reader of this function what the superclass constructor does, and the tricky dependency between needing to set the self.db_config instance variable before calling super().__init__() which then calls _configure_engine which relies on the side effect of self.db_config already being set, and itself having a side effect of setting self._engine, feels like a bit too much reliance on side effects.

at this point i think it'd be cleaner to remove `__init__` and `_configure_engine` from the DBClient abstract base class and instead just make it clear from the typing system that self._engine needs to be set in the constructor (i.e. it has no default value and is not an Optional type and therefore needs to be set) i.e.

```python
class DBClient(ABC):
  # Concrete subclasses need to set the _engine instance var in the constructor.
  # It should not be None.
  _engine: sqlalchemy.engine.Engine
```

then here we can remove all other reliance on side effects e.g.
```suggestion
    def __init__(self, db_config: DBConfig | None = None) -> None:
        if not db_config:
            db_config = get_db_config()
        
        self._engine = self._configure_engine(db_config)

        if db_config.check_connection_on_init:
            check_db_connection()

    def _configure_engine(self, db_config) -> sqlalchemy.engine.Engine:
        # We want to be able to control the connection parameters for each
        # connection because for IAM authentication with RDS, short-lived tokens are
        # used as the password, and so we potentially need to generate a fresh token
        # for each connection.
        #
        # For more details on building connection pools, see the docs:
        # https://docs.sqlalchemy.org/en/13/core/pooling.html#constructing-a-pool
        def get_conn() -> Any:
            return psycopg2.connect(**get_connection_parameters(db_config))
```

lorenyu (approved):
Awesome thanks for improving the platform!

app/src/adapters/db/clients/postgres_client.py
chouinar: Let me see if typing allows that, otherwise I'll just make the init function take in an engine that the subclass is expected to handle.

app/src/adapters/db/clients/postgres_client.py
chouinar: I went ahead and renamed it and moved it to the clients folder to be next to the client itself.

app/src/adapters/db/clients/postgres_client.py
chouinar: Actually was perfectly fine to do that, so that does seem a bit cleaner

PR #153, opened 2023-03-08
Make fixes to the logging setup for testing

chouinar:
## Ticket

TODO

## Changes
Fix a few issues that were occurring when running tests with regards to logging.
1. Test formatting no longer is broken during tests and maintains the user configured values
2. Logs no longer get duplicated for every call to `logging.init`
3. Made logging init into a class that handles entry and exit

## Context for reviewers
todo

## Testing
TODO

Running `make test args="tests -s -vv"` now looks like:
Screen Shot 2023-03-08 at 3 04 37 PM

Which previously looked like:
Screen Shot 2023-03-08 at 3 05 18 PM

### DIFF
--- a/app/src/app.py
+++ b/app/src/app.py
@@ -21,7 +21,7 @@ logger = logging.getLogger(__name__)
 def create_app() -> APIFlask:
     app = APIFlask(__name__)
 
-    src.logging.Log(__package__)
+    src.logging.init(__package__)
     flask_logger.init_app(logging.root, app)
 
     db_client = db.PostgresDBClient()
--- a/app/src/app.py
+++ b/app/src/app.py
@@ -21,8 +21,8 @@ logger = logging.getLogger(__name__)
 def create_app() -> APIFlask:
     app = APIFlask(__name__)
 
-    root_logger = src.logging.Log(__package__).root_logger
-    flask_logger.init_app(root_logger, app)
+    src.logging.Log(__package__)
+    flask_logger.init_app(logging.root, app)
 
     db_client = db.init()
     flask_db.init_app(db_client, app)
--- a/app/src/app.py
+++ b/app/src/app.py
@@ -21,7 +21,7 @@ logger = logging.getLogger(__name__)
 def create_app() -> APIFlask:
     app = APIFlask(__name__)
 
-    root_logger = src.logging.init(__package__)
+    root_logger = src.logging.Log(__package__).root_logger
     flask_logger.init_app(root_logger, app)
 
     db_client = db.init()
--- a/app/src/db/migrations/env.py
+++ b/app/src/db/migrations/env.py
@@ -27,7 +27,7 @@ config = context.config
 logger = logging.getLogger("migrations")
 
 # Initialize logging
-with src.logging.Log("migrations"):
+with src.logging.init("migrations"):
 
     if not config.get_main_option("sqlalchemy.url"):
         uri = make_connection_uri(get_db_config())
--- a/app/src/db/migrations/env.py
+++ b/app/src/db/migrations/env.py
@@ -27,88 +27,88 @@ config = context.config
 logger = logging.getLogger("migrations")
 
 # Initialize logging
-src.logging.init("migrations")
+with src.logging.Log("migrations"):
 
-if not config.get_main_option("sqlalchemy.url"):
-    uri = make_connection_uri(get_db_config())
+    if not config.get_main_option("sqlalchemy.url"):
+        uri = make_connection_uri(get_db_config())
 
-    # Escape percentage signs in the URI.
-    # https://alembic.sqlalchemy.org/en/latest/api/config.html#alembic.config.Config.set_main_option
-    config.set_main_option("sqlalchemy.url", uri.replace("%", "%%"))
+        # Escape percentage signs in the URI.
+        # https://alembic.sqlalchemy.org/en/latest/api/config.html#alembic.config.Config.set_main_option
+        config.set_main_option("sqlalchemy.url", uri.replace("%", "%%"))
 
-# add your model's MetaData object here
-# for 'autogenerate' support
-# from myapp import mymodel
-# target_metadata = mymodel.Base.metadata
-target_metadata = metadata
+    # add your model's MetaData object here
+    # for 'autogenerate' support
+    # from myapp import mymodel
+    # target_metadata = mymodel.Base.metadata
+    target_metadata = metadata
 
-# other values from the config, defined by the needs of env.py,
-# can be acquired:
-# my_important_option = config.get_main_option("my_important_option")
-# ... etc.
+    # other values from the config, defined by the needs of env.py,
+    # can be acquired:
+    # my_important_option = config.get_main_option("my_important_option")
+    # ... etc.
 
+    def include_object(
+        object: sqlalchemy.schema.SchemaItem,
+        name: str,
+        type_: str,
+        reflected: bool,
+        compare_to: Any,
+    ) -> bool:
+        if type_ == "schema" and getattr(object, "schema", None) is not None:
+            return False
+        else:
+            return True
 
-def include_object(
-    object: sqlalchemy.schema.SchemaItem, name: str, type_: str, reflected: bool, compare_to: Any
-) -> bool:
-    if type_ == "schema" and getattr(object, "schema", None) is not None:
-        return False
-    else:
-        return True
+    def run_migrations_offline() -> None:
+        """Run migrations in 'offline' mode.
 
+        This configures the context with just a URL
+        and not an Engine, though an Engine is acceptable
+        here as well.  By skipping the Engine creation
+        we don't even need a DBAPI to be available.
 
-def run_migrations_offline() -> None:
-    """Run migrations in 'offline' mode.
+        Calls to context.execute() here emit the given string to the
+        script output.
 
-    This configures the context with just a URL
-    and not an Engine, though an Engine is acceptable
-    here as well.  By skipping the Engine creation
-    we don't even need a DBAPI to be available.
+        """
 
-    Calls to context.execute() here emit the given string to the
-    script output.
-
-    """
-
-    url = config.get_main_option("sqlalchemy.url")
-    context.configure(
-        url=url,
-        target_metadata=target_metadata,
-        literal_binds=True,
-        dialect_opts={"paramstyle": "named"},
-        include_schemas=False,
-        include_object=include_object,
-        compare_type=True,
-    )
-
-    with context.begin_transaction():
-        context.run_migrations()
-
-
-def run_migrations_online() -> None:
-    """Run migrations in 'online' mode.
-
-    In this scenario we need to create an Engine
-    and associate a connection with the context.
-
-    """
-
-    url = config.get_main_option("sqlalchemy.url", "SQLAlchemy URL not set")
-    connectable = sqlalchemy.create_engine(url)
-
-    with connectable.connect() as connection:
+        url = config.get_main_option("sqlalchemy.url")
         context.configure(
-            connection=connection,
+            url=url,
             target_metadata=target_metadata,
+            literal_binds=True,
+            dialect_opts={"paramstyle": "named"},
             include_schemas=False,
             include_object=include_object,
             compare_type=True,
         )
+
         with context.begin_transaction():
             context.run_migrations()
 
+    def run_migrations_online() -> None:
+        """Run migrations in 'online' mode.
 
-if context.is_offline_mode():
-    run_migrations_offline()
-else:
-    run_migrations_online()
+        In this scenario we need to create an Engine
+        and associate a connection with the context.
+
+        """
+
+        url = config.get_main_option("sqlalchemy.url", "SQLAlchemy URL not set")
+        connectable = sqlalchemy.create_engine(url)
+
+        with connectable.connect() as connection:
+            context.configure(
+                connection=connection,
+                target_metadata=target_metadata,
+                include_schemas=False,
+                include_object=include_object,
+                compare_type=True,
+            )
+            with context.begin_transaction():
+                context.run_migrations()
+
+    if context.is_offline_mode():
+        run_migrations_offline()
+    else:
+        run_migrations_online()
--- a/app/src/logging/__init__.py
+++ b/app/src/logging/__init__.py
@@ -39,15 +39,14 @@ _original_argv = tuple(sys.argv)
 
 class Log:
     def __init__(self, program_name: str) -> None:
-        self.program_name = program_name
-        self.root_logger, self.stream_handler = config.configure_logging()
-        log_program_info(self.program_name)
+        self.stream_handler = config.configure_logging()
+        log_program_info(program_name)
 
-    def __enter__(self) -> logging.Logger:
-        return self.root_logger
+    def __enter__(self) -> None:
+        return None
 
     def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
-        self.root_logger.removeHandler(self.stream_handler)
+        logging.root.removeHandler(self.stream_handler)
 
 
 def log_program_info(program_name: str) -> None:
--- a/app/src/logging/__init__.py
+++ b/app/src/logging/__init__.py
@@ -11,7 +11,8 @@ src.logging.pii for more information.
 Usage:
     import src.logging
 
-    src.logging.init("program name")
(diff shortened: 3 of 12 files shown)


app/src/app.py
```diff
@@ -21,7 +21,7 @@
 def create_app() -> APIFlask:
     app = APIFlask(__name__)
 
-    root_logger = src.logging.init(__package__)
+    root_logger = src.logging.Log(__package__).root_logger
```
chouinar: Probably a better way to do this, I didn't want to do a `with ...` here because we just want it to live forever for the lifespan of an app so don't care about exit being called.

app/src/logging/formatters.py
```diff
@@ -21,7 +22,8 @@ def format(self, record: logging.LogRecord) -> str:
         # logging.Formatter.format adds the `message` attribute to the LogRecord
         # see https://github.com/python/cpython/blob/main/Lib/logging/__init__.py#L690-L720
         super().format(record)
-        return json.dumps(record.__dict__, separators=(",", ":"))
+
+        return json.dumps(record.__dict__, separators=(",", ":"), default=pydantic_encoder)
```
chouinar: This doesn't need to be pydantic, I just know it works.

This solves an issue where certain values can't be serialized. UUIDs and dates in particular fail to serialize properly, which pydantic handles. We can make our own copying what it does, just was fixing it in the most direct way I was aware of

Note that this wasn't an issue in older versions of this code because the `pii_mask` method implicitly made them strings.

app/tests/conftest.py
```diff
@@ -17,7 +17,7 @@
 logger = logging.getLogger(__name__)
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture(scope="session", autouse=True)
```
chouinar, line 20: Before, the `app` fixture was initializing before this was called, this fixes that

app/src/db/migrations/env.py
```diff
@@ -27,88 +27,88 @@
...
-    In this scenario we need to create an Engine
-    and associate a connection with the context.
-
-    """
-
-    url = config.get_main_option("sqlalchemy.url", "SQLAlchemy URL not set")
-    connectable = sqlalchemy.create_engine(url)
-
-    with connectable.connect() as connection:
+with src.logging.Log("migrations"):
```
chouinar: I only tabbed this file over here - git is just confused

app/src/logging/formatters.py
chouinar: If we wanted to do this a bit more directly, something like:

```
def json_encoder(obj: Any) -> Any:

    match obj:
        case str() | int() | float() | bool() | None:
            return obj
        case datetime() | date():
            return obj.isoformat()
        case Decimal():
            if obj.as_tuple().exponent >= 0:
                return int(obj)
            return float(obj)
        case Enum():
            return obj.value
        case UUID():
            return str(obj)
        case _:
            return str(obj)
```
Seems to work - from testing the only things that fall into the last case are exceptions, a class in alembic, and the `exec` issue from auditing. They all get printed semi-readable, for example the exec audit log ends up as:
`"audit.args.code_object":"<code object <module> at 0x1062a3cf0, file \"/Users/michaelchouinard/.pyenv/versions/3.11.0/lib/python3.11/encodings/hex_codec.py\"`

chouinar: I went ahead and changed it, let me know what you think

app/src/logging/__init__.py
```diff
@@ -36,11 +37,17 @@
...
-def init(program_name: str) -> logging.Logger:
-    root_logger = config.configure_logging()
+class Log:
+    def __init__(self, program_name: str) -> None:
+        self.program_name = program_name
+        self.root_logger, self.stream_handler = config.configure_logging()
+        log_program_info(self.program_name)
 
-    log_program_info(program_name)
-    return root_logger
```
lorenyu: Nit: instead of making a class, I was hoping to keep it a little simpler/lower-tech by using the [contextmanager decorator](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) i.e.

```
from contextlib import contextmanager

@contextmanager
def init(program_name: str) -> logging.Logger:
    root_logger, stream_handler = config.configure_logging()
    log_program_info(program_name)
    try:
        yield root_logger
    finally:
        root_logger.removeHandler(stream_handler)
```

I think a class-based contextmanager is helpful when it gets more complex but in this case I think the cleanup step we're trying to do is relatively simple so I think it's a bit easier to read when it's all in one function like this.

Going further, I actually think that it would be cleaner to leave the `init` function alone and move all the context manager logic to the configure_logging function, since that function is the one that sets everything up and therefore it's the function that should know what needs to be cleaned up. In other words, right now we happen to know that the console handler needs to be removed from the root logger, but that's leaking the implementation details of the configure_logging function. So we'd end up with something like:

**src/logging/__init__.py**
```python
def init(program_name) -> typing.ContextManager[None]:
  loggingcontextmanager = config.configure_logging()
  log_program_info(program_name)
  return loggingcontextmanager
```

**src/logging/config.py**
```python
from contextlib import contextmanager

@contextmanager
def configure_logging()
    ...
    logging.root.addHandler(console_handler)
    ...
    
    try:
        yield
    finally:
        logging.root.removeHandler(console_handler)
```

and we can eliminate the return values from both of these functions too.

app/tests/conftest.py
lorenyu: Good catch, could you add that as a comment?

app/src/app.py
lorenyu: Two thoughts. First, if we want to preserve the context manager abstraction and not rely on the fact that there's a class instance variable called root_logger (see my comment below about just using a `@contextmanager` decorator rather than creating the `Log` class), then we can do

```python
root_logger = src.logging.init(__package__).__enter__()
```

But my second thought is that we only care that `src.logging.init()` initializes the logging, we don't _really_ care about the return value, and this is the only place in the code where we use the return value from init anyways, so my instinct is to remove the return value from `src.logging.init` algother, and change this code to just

```
src.logging.init(__package__)
flask_logger.init_app(logging.root, app)
```

app/src/logging/formatters.py
```diff
@@ -6,22 +6,54 @@
...
+        case UUID():
+            return str(obj)
+        case Exception():
+            return str(obj)
+        case set():
+            # The JSON library will handle
+            # iterating over and potentially calling this again
+            return list(obj)
+        case _:
+            return str(obj)
```
lorenyu: How much of this code is equivalent to the default behavior? I was looking at the reference implementation of extending the JSON encoder in the docs, https://docs.python.org/3/library/json.html, and I like that in their example implementation of "ComplexEncoder" they only need to define the special case and default to [json.JSONEncoder's default method](https://docs.python.org/3/library/json.html#json.JSONEncoder.default) otherwise, so I'm wondering if we only need to implement for things like `Enum` that are custom

lorenyu (commented):
Thanks for finding and addressing these issues!

app/src/logging/formatters.py
chouinar: Anything besides str, int, float, bool, None, and lists of those types doesn't work without this change.

I could trim it a bit as decimal/uuid/exception all would be served by the fallback case as they're the same, I just wanted to be explicit that they're handled - could be a comment.

I made a decision here that rather than fail if we didn't have a converter, we'd defer to `str()`. Since this is only for logging, and most systems that would ingest the logs would only support basic types anyways, I feel that is fine. The problem with the way the default approach works currently is that it fails silently unless you go dig through the logs.

lorenyu: I see, that's a good point, could you add that as a comment (that we want to fallback to `str()` rather than failing?)

For my own understanding, in the default approach, when it fails, what does it show in the logs? Is it an empty string or does it look like some sort of exception?

app/src/logging/__init__.py
chouinar: I went with the class approach as I could not get this way to work.

The issue is, we want the app itself to only go into `__enter__` but never exit. If I specify in app.py `src.logging.init(..).__enter__()` it goes into the finally block (when using it as `with ..init(..)` it's fine).

Don't think we can use a context manager for this unless you've got another idea?

chouinar: I tidied it up some by removing the root logger needing to be passed around as suggested on another thread.

Let me know if the class approach seems better with that, or if you can think of another way to make it work.

app/src/logging/formatters.py
chouinar: You get a very long stack trace like:
```
TypeError: Object of type UUID is not JSON serializable
Call stack:
  ...
  File "/Users/michaelchouinard/workspace/template-application-flask/app/src/api/users/user_routes.py", line 53, in user_get
    logger.info("Successfully fetched user", extra=get_user_log_params(user))
Message: 'Successfully fetched user'
Arguments: ()
```
And as far as I can tell, that log message doesn't end up being created.

app/src/logging/__init__.py
lorenyu: Ah I see, doing some experimentation, it looks like if you do

```python
def create_app(...):
  ...
  contextmgr = src.logging.init(...).__enter__()
  ...
```

when the function returns and the `contextmgr` variable goes out of scope, it automatically calls __exit__(), or in your case, there never is a variable so it immediately calls __exit__().

i guess the class is the only way i can think of. maybe we can think of a different name for it , i think of Log as synonymous with Logger. i'll re-review and see if i can think of a different name

app/src/logging/config.py
```diff
@@ -46,6 +47,7 @@ def configure_logging() -> logging.Logger:
     formatter = get_formatter(config)
     console_handler.setFormatter(formatter)
     console_handler.addFilter(pii.mask_pii)
+    logging.root.removeHandler(console_handler)
```
lorenyu: should we delete this line? it looks like a no-op. i don't see how the newly created console_handler could already exist on the root.

app/src/logging/__init__.py
```diff
@@ -36,11 +37,16 @@
 _original_argv = tuple(sys.argv)
 
 
-def init(program_name: str) -> logging.Logger:
-    root_logger = config.configure_logging()
+class Log:
```
lorenyu: i thought of some other names:
- `LoggingContext`
- `ConfiguredLogging`
- `InitializedLogging`
- `ConfiguredLoggingContext`
- `InitializedLoggingContext`
- `ConfigureLogging`
- `ConfigureLoggingContext`

i have a slight preference for keeping it simple and just appending Context and calling it `LoggingContext`

lorenyu: for type hints could we have Log implement ContextManager e.g.

```python
from typing import ContextManager

class LoggingContext(ContextManager[None]):
```

app/src/logging/__init__.py
```diff
@@ -36,11 +37,16 @@
...
+        self.stream_handler = config.configure_logging()
+        log_program_info(program_name)
 
-    log_program_info(program_name)
-    return root_logger
+    def __enter__(self) -> None:
+        return None
+
+    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
+        logging.root.removeHandler(self.stream_handler)
```
lorenyu: could we consolidate this class into `config.py` and consolidate the configure_logging function into a method on this class so that `logging.root.addHandler` and `logging.root.removeHandler` are at the same level (i.e. in the same class)

app/src/logging/__init__.py
```diff
@@ -11,7 +11,8 @@
 Usage:
     import src.logging
 
-    src.logging.init("program name")
+    with src.logging.Log("program name"):
```
lorenyu: Nit: Purely stylistically it'd be nice to be consistent with the pattern that there's an all lower case function that developers call that return a context manager, rather than instantiating the context manager itself. e.g.

`engine.connect()` returns `Connection`, `open` returns a `File`, etc., so could we still have a wrapper function like `init`/`init_logging` or `configure`/`configure_logging` that just returns the context manager? i.e.

```python
def init(program_name: str) -> LoggingContext:
    return LoggingContext(program_name)
```

app/src/logging/formatters.py
```diff
@@ -6,22 +6,69 @@
...
 
+def json_encoder(obj: Any) -> Any:
+    """
+    Handle conversion of various types when logs
+    are serialized into JSON. If not specified
+    will attempt to convert using str() on the object
+    """
+
+    # This is the equivalent of many isinstance(obj, <type>) checks
+    match obj:
```
lorenyu: since this is such a low level function that can be called dozens of times per log record, and potentially hundreds of thousands of times for a script that processes a lot of data, i'm a little more concerned about performance.

my understanding of `match` is that it essentially is a much more readable way of doing if/then statements, but it's still scales linearly in the number of statements, whereas a lookup table would be constant time (https://betterprogramming.pub/pythons-match-case-is-too-slow-if-you-don-t-understand-it-8e8d0cf927d)

since we're not doing any fancy pattern matching, can we instead index on the type of the object e.g.

```python

identity = lambda x: x
default_encoder = str
ENCODERS_BY_TYPE = defaultdict(lambda: default_encoder, {
    "str": identity,
    "int": identity,
    "float": identity,
    "bool": identity,
    "int": identity,
    "None": identity,
    "datetime": lambda d: d.isoformat(),
    "date": lambda d: d.isoformat(),
    "Decimal": str,
    "UUID": str,
    "Exception": str,
    "Enum": lambda e: e.value,
    "set": lambda s: list(s),
})


def json_encoder(obj):
    encode = ENCODERS_BY_TYPE[type(obj)]
    return encode(obj)
```

i think this will have more predictable constant time performance

app/tests/src/logging/test_flask_logger.py
```diff
@@ -12,9 +12,14 @@
...
     logger = logging.getLogger("src")
+    before_level = logger.level
+
     logger.setLevel(logging.DEBUG)
-    logger.addHandler(logging.StreamHandler(sys.stdout))
-    return logger
+    handler = logging.StreamHandler(sys.stdout)
+    logger.addHandler(handler)
+    yield logger
+    logger.setLevel(before_level)
```
lorenyu, line 21: nice

lorenyu (approved):
Looks good, left some final comments

app/src/logging/formatters.py
chouinar: Good point on the lookup > match performance, I'll just steal exactly what pydantic_encoder did for that

lorenyu: oh cool, looks like they did something [very similar](https://github.com/pydantic/pydantic/blob/main/pydantic/json.py#L45-L70) and LoL they even came up with the same variable name haha that's a funny coincidence

app/src/logging/config.py
```diff
@@ -26,41 +30,77 @@ class Config:
...
-    logging.getLogger("sqlalchemy.dialects.postgresql").setLevel(logging.INFO)
-
-    return console_handler
+
+    def __init__(self, program_name: str) -> None:
+        self._configure_logging()
+        log_program_info(program_name)
+
+    def __enter__(self) -> None:
+        return None
```
lorenyu: neither here nor there but i think this is equivalent to `pass` which _slightly_ communicates the intention of "no-op" a little better

app/src/logging/formatters.py
```diff
@@ -11,53 +11,64 @@
...
 from enum import Enum
-from typing import Any
+from typing import Any, Callable, Type
 from uuid import UUID
 
 import src.logging.decodelog as decodelog
 
 
+# identity returns unmodified value
+def identity(obj: Any) -> Any:
```
lorenyu: what i'm about to say also doesn't matter at all to me, but if you _wanted_ to be more rigorous with typing you'd make this a generic, since the type of the return is the same as the parameter type, so it'd be:

```python
from typing import TypeVar
T = TypeVar("T")
def identity(obj: T) -> T:
    return obj
```

app/src/logging/config.py
```diff
@@ -26,41 +30,77 @@ class Config:
...
+
+    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
+        # Remove the console handler to stop logs from being sent to stdout
```
lorenyu, line 69: nit: is it worth explaining a little bit more the value of cleaning up? e.g. "This is useful in the test suite, since through the testing session the logger may be configured multiple times, so we want to clean up after each time" or something like that (feel free to make it less wordy)

lorenyu (approved):
Looks great!

PR #155, opened 2023-03-14
Fix error when PII appears in interpolated logs

### LINKED ISSUE #154: Logger errors when PII appears in interpolation args
error when trying to do something like ` logger.info("%s %s", "text", "123456789")`

see 🔒 [slack thread](https://nava.slack.com/archives/C03G1SWD9H7/p1678807142681979)

lorenyu:
## Ticket

Resolves #154 

## Changes
* Mask PII interpolation args separate from extra

## Context
Previously, something like `logger.info("%s %s", "text", "123456789")` would cause a logger error
since the mask_pii logic would replace the LogRecord's `args` tuple with the string '("text", "*********")',
which would break the log interpolation logic. This PR fixes that bug and adds some regression tests.

## Testing
Added tests that failed before the fix and pass after the fix

### DIFF
--- a/app/src/logging/pii.py
+++ b/app/src/logging/pii.py
@@ -34,9 +34,22 @@ from typing import Any, Optional
 
 
 def mask_pii(record: logging.LogRecord) -> bool:
+    # Loop through all entries in the record's __dict__
+    # attribute and mask any things that look like PII.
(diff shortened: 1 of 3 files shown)


chouinar (approved):
LGTM! Thanks for fixing this!

PR #158, opened 2023-03-21
Replace Docker Desktop with Colima in prereqs

### LINKED ISSUE #157: Remove Docker Desktop from instructions
# Context

Docker Desktop isn't free anymore so we should default devs to a free tool that supports docker CLI

lorenyu:
## Ticket

Resolves #157 

## Changes

see title

## Context

Docker Desktop is no longer universally free so as a default we are going to recommend a free alternative. Colima so far seems to be a good alternative.

## Testing
n/a

### DIFF
--- a/docs/app/getting-started.md
+++ b/docs/app/getting-started.md
@@ -21,7 +21,17 @@ A very simple [docker-compose.yml](/docker-compose.yml) has been included to sup
 
 3. If you are using an M1 mac, you will need to install postgres as well: `brew install postgresql` (The psycopg2-binary is built from source on M1 macs which requires the postgres executable to be present)
 
-4. You'll also need [Docker Desktop](https://www.docker.com/products/docker-desktop/)
+4. You'll also need a container runtime ([Colima](https://github.com/abiosoft/colima) and Docker CLI. You can install both of these with:
+
+   ```bash
+   brew install colima docker
+   ```
+
+   After you finish installing, you'll want to start the docker daemon with
+
+   ```bash
+   colima start
+   ```
 
 ## Setup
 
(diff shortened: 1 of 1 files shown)


docs/app/getting-started.md
```diff
@@ -21,7 +21,17 @@ A very simple [docker-compose.yml](/docker-compose.yml) has been included to sup
...
+4. You'll also need a container runtime ([Colima](https://github.com/abiosoft/colima) and Docker CLI. You can install both of these with:
+
+   ```bash
+   brew install colima docker
+   ```
+
+   After you finish installing, you'll want to start the docker daemon with
+
+   ```bash
+   colima start
```
chouinar, line 33: Some thoughts.

We've had to use colima on the NJ project in order to run an Oracle DB locally as the M1 Mac chip isn't supported. Colima is able to use some system tools to emulate the chip by doing `colima start --arch x86_64`. This has a pretty substantial performance penalty, and I'm curious if just plain colima has any as well. For us, it's very noticeable in the docker build + tests. I'd guess for just plain colima, it's not.

MIght be worth documenting a few additional params like the memory and cores you give to colima as the defaults are low. Also what exactly colima is as it's a docker context, and docker only allows a single global context on your machine which has been painful for us (we'd love to just run the Oracle image with the emulation locally, but we have to do the API too).

lorenyu: My lack of Docker expertise is coming out here, but I don't really understand much of what you're saying (not sure what a docker context is for example), and I'm not sure what params to provide colima or whether those params should be defaulted across the organization or whether we should just provide a reference to colima docs for people to tune to their own needs and computer resources.

I haven't noticed a big performance penalty so far, and from the online threads I've come across some even claim it's faster than docker desktop's engine.

lorenyu: If there are some learnings that you have that you want to add, do you want to jump on this branch and add those things? Otherwise I'm not really sure what additional guidance to give. I'm just trying to meet the MVP needs of getting people off Docker Desktop

lorenyu: I'm actually running into some issues with colima. For some reason the image I built with `docker build` doesn't run in the cloud with the template-infra repo. The one with docker buildx build works. So I'm going to close this for now

PR #159, opened 2023-03-23
Update LICENSE to mention Nava

bbemis017:
## Ticket

n/a

## Changes
- Updates the LICENSE to mention `Nava PBC` on the copyright line.

## Context for reviewers

I left the [yyyy] brackets there as a reminder that the year should be updated when the template is copied. But let me know if that doesn't seem right.

## Testing
n/a

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] Nava PBC
+   Copyright 2022 Nava PBC
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright [yyyy] Nava PBC
```
lorenyu:
```suggestion
   Copyright 2022 Nava PBC
```

lorenyu (approved):
Thanks!

lorenyu:
@bbemis017 sorry I forgot to read your PR description first.

I think we should include both the year and the org, because this license is for the template itself. The license file wouldn't be copied to project repos.

PR #160, opened 2023-04-19
Fix Makefile CI functionality

Nava-JoshLong:
## Ticket

{TICKET_LINK}

## Changes

- Added missing `SHELL` line in Makefile
- Edited file paths in `CI` flag paths to match expected path
- Added comments to explain

## Context for reviewers

The flags in the `ifdef CI` section will run in the GH workflow, and need to have the SHELL flag command defined so that they can show errors in the files changed section, as well as fail as expected because it is "recasting" the error to a format GH can ingest

## Testing

Adding a function that will fail linting, without this PR will continue in the workflow

### DIFF
--- a/app/Makefile
+++ b/app/Makefile
@@ -15,16 +15,16 @@ DECODE_LOG := 2>&1 | python3 -u src/logging/util/decodelog.py
 # Required for CI flags below to work properly
 SHELL = /bin/bash -o pipefail
 
-# A few commands need adjustments if they're run in CI, specify those here
-# Make sure the path of the file mirrors the path to the app, or these flags will
-# not work as expected in GitHub
-## FLAKE8_FORMAT will show any findings in the "Files Changed" tab of the PR
-## MYPY_POSTPROC will show any findings in the "Files Changed" tab of the PR
+# The APP_DIR variable is the path from the root of the repository to this Makefile.
+# This variable is used to display errors from Flake8 and MyPy in the 'Files Changed'
+# section of a pull request. If this is set to the incorrect value, you won't be able
+# to see the errors on the correct files in that section
+APP_DIR = app
 ifdef CI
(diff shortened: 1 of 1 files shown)


chouinar (approved):

app/Makefile
```diff
@@ -12,13 +12,19 @@ APP_NAME := main-app
...
 # A few commands need adjustments if they're run in CI, specify those here
-# TODO - when CI gets hooked up, actually test this.
+# Make sure the path of the file mirrors the path to the app, or these flags will
+# not work as expected in GitHub
+## FLAKE8_FORMAT will show any findings in the "Files Changed" tab of the PR
+## MYPY_POSTPROC will show any findings in the "Files Changed" tab of the PR
 ifdef CI
  DOCKER_EXEC_ARGS := -T -e CI -e PYTEST_ADDOPTS="--color=yes"
- FLAKE8_FORMAT := '::warning file=src/%(path)s,line=%(row)d,col=%(col)d::%(path)s:%(row)d:%(col)d: %(code)s %(text)s'
+ FLAKE8_FORMAT := '::warning file=app/%(path)s,line=%(row)d,col=%(col)d::%(path)s:%(row)d:%(col)d: %(code)s %(text)s'
```
lorenyu: Just a small nit could we define `APP_DIR := app` above and then do
```suggestion
 FLAKE8_FORMAT := '::warning file=$(APP_DIR)/%(path)s,line=%(row)d,col=%(col)d::%(path)s:%(row)d:%(col)d: %(code)s %(text)s'
```
at some point I want to make the `app` a configurable thing to more easily support multiple apps per infra setup, and that will make it easier to catch than having app buried in the string

lorenyu (approved):
Thanks for fixing!

app/Makefile
Nava-JoshLong: Updated the file to use the variable version of this and changed the comment to explain what the variable is and why it's needed

PR #161, opened 2023-05-08
Experiment: multi-environment configuration in Python

jamesbursa:
## Changes

- Add new `src.config` module that defines config for each environment entirely in Python.
- Modify startup code to load that config and pass it to each module that needs it.
- Update tests to match (remove most monkeypatching of environ).

## Context for reviewers

Start with default.py and local.py to understand how this works.

Goals of this approach:

1. Developer friendly with minimal pitfalls - find out at test time if a config option is not set or invalid type for any environment
2. Almost all settings are in python code instead of split between infrastructure (terraform / env vars) and code
3. Still possible to override a setting from an env var if needed (for secrets or runtime overrides)
4. Same source of settings when running in docker-compose, native, within IDE, or running the test suite
5. Tests aren't accidentally affected by env vars that you have locally

## Testing

Note: some tests are still broken.


app/src/__main__.py
```diff
@@ -16,23 +18,26 @@
...
-    app_config = AppConfig()
+    config = src.config.load(environment_name=os.getenv("ENVIRONMENT", "local"), environ=os.environ)
 
-    app = src.app.create_app()
+    app = src.app.create_app(config)
+    logger.info("loaded configuration", extra={"config": config})
 
-    environment = app_config.environment
+    all_configs = src.config.load_all()
+    logger.info("loaded all", extra={"all_configs": all_configs})
```
jamesbursa: This code should move to a unit test, where it checks that each one is an instance of `RootConfig`.

app/src/config/default.py
```diff
@@ -0,0 +1,26 @@
...
+def default_config():
+    return RootConfig(
+        app=AppConfig(),
+        database=PostgresDBConfig(),
+        logging=LoggingConfig(
+            format="json",
+            level="INFO",
+            enable_audit=True,
+        ),
+    )
```
chouinar: Would this require any process that runs to be able to load all environment variables? For example, if I have a config for connecting to a separate service, and it's only used in a few select processes, would I put that here in the default config? Let's assume that I don't want to set defaults in the config and want it to error if any fields aren't set.

app/src/config/env/local.py
```diff
@@ -0,0 +1,16 @@
...
+import pydantic.types
+
+from .. import default
+
+config = default.default_config()
+
+config.database.password = pydantic.types.SecretStr("secret123")
+config.database.hide_sql_parameter_logs = False
+config.logging.format = "human_readable"
+config.logging.enable_audit = False
```
chouinar: Would there be a way to have this check for a file named `local_overrides.py` and use that? I'm thinking about local testing where I want to tinker with environment variables, but not check them in - especially when working with passwords or keys to access services. We could add this `local_overrides.py` file to the `.gitignore`.

Adding something like this to the end of this file:

```py
if os.path.exists("local_overrides.py"):
     importlib.import_module("local_overrides") # whatever the right path/params are
```

app/src/app_config.py
```diff
@@ -2,7 +2,7 @@
 
 
 class AppConfig(PydanticBaseEnvConfig):
-    environment: str
+    environment: str = "unknown"
```
KevinJBoyer, line 5: Maybe `environment: Optional[str] = None`? That might be annoying with the linter though

app/src/util/env_config.py
```diff
@@ -1,16 +1,30 @@
...
+    def override_from_environment(self, environ, prefix=""):
+        """Recursively override field values from the given environment variable mapping."""
+        for name, field in self.__fields__.items():
+            if field.is_complex():
+                # Nested models must be instances of this class too.
+                getattr(self, name).override_from_environment(environ, prefix=name + "_")
+                continue
+
+            env_var_name = field.field_info.extra.get("env", prefix + name)
+            for key in (env_var_name, env_var_name.lower(), env_var_name.upper()):
```
KevinJBoyer, line 32: Might be nice to be stricter here and require a certain casing, avoid letting someone accidentally set the environment variable in lowercase and then not understand why it's not being overridden if it's also set in uppercase elsewhere.

app/src/util/env_config.py
```diff
@@ -1,16 +1,30 @@
...
+class PydanticBaseEnvConfig(BaseModel):
+    """Base class for application configuration.
+
+    Similar to Pydantic's BaseSettings class, but we implement our own method to override from the
+    environment so that it can be run later, after an instance was constructed."""
 
-class PydanticBaseEnvConfig(BaseSettings):
     class Config:
-        env_file = env_file
+        validate_assignment = True
```
KevinJBoyer, line 18: Could require a prefix option here too to provide some namespacing for the various configs, e.g., have the PostgresConfig environment variables all start with a `PG_DB_` prefix.

KevinJBoyer: (The implementation on PFML does this)

PR #168, opened 2023-06-02
removed v1 from url to test in getting started doc

SammySteiner:


lorenyu (approved):
Thanks for catching this!

PR #172, opened 2023-06-09
Integrate with infra template

lorenyu:
## Ticket

Work related to:
- https://github.com/navapbc/template-infra/issues/216
- https://github.com/navapbc/template-application-flask/issues/171

Resolves https://github.com/navapbc/template-application-flask/issues/171

## Changes
Flask template changes:
- Add release-build target to Makefile to meet [infra template application requirements](https://github.com/navapbc/template-infra/blob/main/template-only-docs/application-requirements.md)
- Update DB environment variable names to match what infra template provides
- Remove load_local_env_vars so that cloud service doesn't use local env vars (rather than updating infra template to set `ENVIRONMENT` env var which application previously relied on). This makes the application closer to a true 12-factor app.
- Update get_connection_parameters to generate new IAM auth token if DB_PASSWORD isn't set

Other changes:
- Update docker-compose main-db container's environment configuration to use minimum env vars rather than loading the full local.env file
- Ignore DL3008
- Update certifi to patch https://github.com/advisories/GHSA-43fp-rhv2-5gv8
- Update cryptography to patch https://github.com/advisories/GHSA-5cpq-8wj7-hf2v
- Update debian packages to fix vulnerability scan errors
- Add production release stage to Dockerfile to fix vulnerability scan errors
- Update app's make release-build to target the release stage

(No infra template related changes)

## Context for reviewers
This PR demonstrates a working integration between the Flask template and the Infra template. Some changes needed to be made to the Flask template in order to make this work.

First, on the `main` branch, I ran through the template-infra setup instructions to set up the infra resources. But after running through the instructions the service wasn't able to be healthy. So I started making changes to the application on this branch to make it work.

Note that this PR might break running the app directly on the host machine, but I plan on solving that later. I think it's higher priority to get the integration/compatibility with the infra template.

## Testing
I made these changes in a new platform-test-flask repo that integrates the flask template and infra template. See: https://github.com/navapbc/platform-test-flask/pull/1

Copied the testing notes from that PR here:

-----

### After making changes to the application on this branch, I manually deployed the application from this branch using the [deploy github action workflow](https://github.com/navapbc/platform-test-flask/actions/workflows/cd.yml) 
image

### Then I verified the app's health checks are working and that the logs look healthy
image

### Finally I checked that the [app's swagger endpoint](http://app-dev-1876508187.us-east-1.elb.amazonaws.com/docs) is loading properly
image

You can see for yourself that it's up and running! http://app-dev-1876508187.us-east-1.elb.amazonaws.com/docs

PR #176, opened 2023-06-14
Add db-migrate to container path to meet infra template requirements

### LINKED ISSUE #175: Integrate with infra migrations
As part of https://github.com/navapbc/template-infra/issues/127 we need to update the flask template to meet the application requirements for the infra template

lorenyu:
## Ticket

Resolves #175

## Changes
* Add db-migrate executable command to application container PATH
* Use consistent method for creating db connection between migrations and application
* Remove code for make_connection_uri
* Remove support for offline migrations

## Context for reviewers
The infra template PR https://github.com/navapbc/template-infra/pull/316/files adds application-requirements that require application container to have a "db-migrate" command available in the PATH. This PR updates template-application-flask to do that.

The previous migrations code connects to the database using a different method than what the application uses, which means that the migrations code does not have the functionality of connecting to the DB using IAM auth, which is a requirement for the infra template. In this change, we make the migrations code in env.py use the existing PostgresDBClient class and remove the extraneous make_connection_uri class.

As part of the change, we also remove support for offline migrations, which is not needed in general. This simplifies the code paths in env.py and reduces maintenance burden.

## Testing
Locally ran `make release-build` then ran `docker -it <image hash> db-migrate`
image

The migration obviously didn't succeed since I didn't pass any necessary env vars, but the db-migrate command was found.

### DIFF
--- a/app/src/db/migrations/env.py
+++ b/app/src/db/migrations/env.py
@@ -10,9 +10,7 @@ from alembic import context
 # See database migrations section in `./database/database-migrations.md` for details about running migrations.
 sys.path.insert(0, ".")  # noqa: E402
 
-
-import src.adapters.db as db
-from src.adapters.db.clients.postgres_config import get_db_config  # noqa: E402 isort:skip
+import src.adapters.db as db  # noqa: E402 isort:skip
 from src.db.models import metadata  # noqa: E402 isort:skip
 import src.logging  # noqa: E402 isort:skip
 
--- a/app/src/db/migrations/env.py
+++ b/app/src/db/migrations/env.py
@@ -3,6 +3,7 @@ import sys
 from typing import Any
 
 import sqlalchemy
+
 from alembic import context
 
 # Alembic cli seems to reset the path on load causing issues with local module imports.
--- a/app/src/db/migrations/env.py
+++ b/app/src/db/migrations/env.py
@@ -1,19 +1,12 @@
 import logging
-import sys
 from typing import Any
 
+import alembic.context as context
 import sqlalchemy
 
-from alembic import context
-
-# Alembic cli seems to reset the path on load causing issues with local module imports.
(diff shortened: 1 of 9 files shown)


app/src/db/migrations/env.py
```diff
@@ -1,24 +1,12 @@
...
 
-# Alembic cli seems to reset the path on load causing issues with local module imports.
-# Workaround is to force set the path to the current run directory (top level src folder)
-# See database migrations section in `./database/database-migrations.md` for details about running migrations.
-sys.path.insert(0, ".")  # noqa: E402
-
-# Load env vars before anything further
-from src.util.local import load_local_env_vars  # noqa: E402 isort:skip
-
-load_local_env_vars()
```
chouinar, line 16: How do the env vars get loaded locally? I thought that was done when running Flask commands, but this doesn't look to run via Flask.

chouinar (approved):
LGTM! We'd done something similar for using the DB client on NJ instead of the `make_connection_uri` method

app/src/db/migrations/env.py
lorenyu: good question. i have to confess that was prioritizing getting database integration working, so i intentionally let running on host machine break. i was gonna work on that soon with https://github.com/navapbc/template-application-flask/issues/136

i'm currently thinking of having developers install direnv, and having an `app/.envrc` file that looks like

```
source local.env
```

and then whenever you `cd` into `app` direnv will automatically get all the env vars from local.env into your shell, and when you `cd` out of the directory direnv will automatically unload those env vars.

the benefit of this is that local developers can also override vars in `local.env` without risking checking those in by adding the overrides to `.envrc` e.g.

```
source local.env
export DB_HOST=localhost
```

lorenyu: (I actually probably already broke running host machine with PR #172 , so it was probably already broken before this PR)

PR #178, opened 2023-06-20
Update database management documentation

### LINKED ISSUE #47: Document instructions for local database management
# Context

the database commands in the makefile could use more comments and some additional documentation in the docs folder to explain how to use them and when to use them

i also couldn't find any instructions on how to start the database from scratch i.e. deleting the volume, which would be useful especially for template development

kylegunby:
## Ticket
Resolves [#47](https://github.com/navapbc/template-application-flask/issues/47)

## Changes
- Renamed the database-migrations doc to `database-management`
- Added documentation for starting, stopping, and cleaning the database
- Removed unused deployment links in database-management document
- Updated `db-recreate` command in Makefile to call `clean-volumes` instead of `clean-docker-volumes`

## Context for reviewers
Some database documentation was missing. I believe it makes more sense to add it to the database-migrations document and rename to database-management.

## Testing
N/A

### DIFF
--- a/docs/app/database/database-management.md
+++ b/docs/app/database/database-management.md
@@ -11,8 +11,9 @@
 ## Basic operations
 ### Initialize
 
-To start a local Postgres database container in a detached state and run any 
-pending migrations, run `make init-db`. Running `make init` will also call `init-db` automatically.
+To start a local Postgres database container in a detached state and run any
+pending migrations, run `make init-db`. During initial setup, `init-db` is called
+automatically when running `make init`.
 
 ### Start
 
@@ -31,7 +32,9 @@ To clean the database, use the following command:
 make db-recreate
 ```
 
-This will remove _all_ docker project volumes, rebuild the database volume, and run all pending migrations. Once completed, only the database container will be running. Simply run `make start` to bring up all other project containers.
+This will remove _all_ docker project volumes, rebuild the database volume, and 
+run all pending migrations. Once completed, only the database container will be 
+running. Simply run `make start` to bring up all other project containers.
 
 ## Running migrations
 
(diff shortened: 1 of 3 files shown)


docs/app/database/database-management.md
lorenyu, line 1: Since you renamed the file can you search for all links to this file? I think there's at least [one in the README](https://github.com/navapbc/template-application-flask/blob/2d389534c03fc1db87e0e6162bf9010df4321b0e/docs/app/README.md?plain=1#L41)

docs/app/database/database-management.md
```diff
@@ -1,11 +1,40 @@
...
 - [Creating new migrations](#creating-new-migrations)
 - [Multi-head situations](#multi-head-situations)
-- [Deployment](#deployment)
-  - [Removing a column](#removing-a-column)
-  - [Removing a table](#removing-a-table)
+
+## Basic operations
+### Initialize
+
+When setting up the environment, `make init` will take care of building the database and running a few basic migrations. Should there be a situation where this process needs to be repeated or ran separately, the following command will accomplish this:
```
lorenyu: > Should there be a situation where this process needs to be repeated or ran separately

nitpick: i think there's ambiguity in the phrase "this process". when i read it, it sounds like the "this process" refers to setting up the environment, so it leaves me wondering why not just re-run `make init`. can you make it a bit clearer? i defer to how you want to word it, but one idea is something like

> Run `make init-db` to start up a local Postgres database container in a detached state and run any pending migrations. Running `make init` will also automatically calls `make init-db`.

or something like that 🤷

lorenyu (approved):
Looks good! Just a couple nitpick comments

docs/app/database/database-management.md
kylegunby: That's a good catch. I've reworded it to point out that it's ran automatically during setup so there is no confusion.

PR #179, opened 2023-06-22
Add basic index route

### LINKED ISSUE #177: Make site index ("/") return something
Currently the site index "/" i.e. the "homepage" returns 404, even if the app is up and running. This is unintuitive since the homepage of a site should always return _something_. Moreover, that assumption is what the infra e2e tests rely on, so the fact that the homepage returns 404 causes infra e2e tests to fail (see this example run in platform-test-flask repo https://github.com/navapbc/platform-test-flask/actions/runs/5324396511/jobs/9643634571)

This ticket is to add some sort of response to the index. To keep it simple it could just be a simple title, description, and link to the "/docs" swagger docs endpoint.

kylegunby:
## Ticket

Resolves [#177](https://github.com/navapbc/template-application-flask/issues/177)

## Changes
Added and index route with basic html including a link to the swagger docs endpoint

## Context for reviewers
Previously the index route returned a 404 causing issues with e2e tests.

## Testing
Added `app/tests/src/route/test_index_route.py`

## Screenshots
Screenshot 2023-06-22 at 1 48 43 PM

### DIFF
--- a/app/openapi.generated.yml
+++ b/app/openapi.generated.yml
@@ -10,33 +10,6 @@ tags:
 - name: Health
 - name: User
 paths:
-  /:
-    get:
-      parameters: []
-      responses:
-        '200':
-          content:
-            application/json:
-              schema:
-                type: object
-                properties:
-                  message:
-                    type: string
-                    description: The message to return
-                  data: {}
-                  status_code:
-                    type: integer
-                    description: The HTTP status code
-                  warnings:
-                    type: array
-                    items:
-                      $ref: '#/components/schemas/ValidationError'
-                  errors:
-                    type: array
-                    items:
-                      $ref: '#/components/schemas/ValidationError'
-          description: Successful response
-      summary: Index
   /health:
     get:
       parameters: []
--- a/app/openapi.generated.yml
+++ b/app/openapi.generated.yml
@@ -10,6 +10,33 @@ tags:
 - name: Health
 - name: User
 paths:
+  /:
+    get:
+      parameters: []
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  message:
+                    type: string
+                    description: The message to return
+                  data: {}
+                  status_code:
+                    type: integer
+                    description: The HTTP status code
+                  warnings:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/ValidationError'
+                  errors:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/ValidationError'
+          description: Successful response
+      summary: Index
   /health:
     get:
       parameters: []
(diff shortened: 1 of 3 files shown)


app/src/app.py
```diff
@@ -29,6 +29,8 @@ def create_app() -> APIFlask:
 
     configure_app(app)
     register_blueprints(app)
+    build_index_route(app)
```
lorenyu: nit: maybe call it `register_index` or `register_index_route` to be consistent with `register_blueprints`

app/openapi.generated.yml
```diff
@@ -10,6 +10,33 @@ tags:
 - name: Health
 - name: User
 paths:
+  /:
+    get:
+      parameters: []
+      responses:
+        '200':
+          content:
+            application/json:
```
lorenyu: weird, why did apiflask generate this as application/json? is it a bug/unsupported functionality in apiflask or did we need to configure something?

lorenyu (commented):
Great work! But I don't think testing should be "N/A" here. How did you know this endpoint works? Can you add a screenshot of what the page looks like when you view it running on localhost?

app/openapi.generated.yml
kylegunby: I saw this and was wondering the same thing. I'll take a look at the documentation and see why it did that.

kylegunby:
> Great work! But I don't think testing should be "N/A" here. How did you know this endpoint works? Can you add a screenshot of what the page looks like when you view it running on localhost?

That's a good point. I will add a screenshot. I can also add a quick test in `tests/src/routes` to be safe.

app/openapi.generated.yml
chouinar: Having looked at how the generation works before, you'll need to look in apispec: https://apispec.readthedocs.io/en/latest/ as it is what generates the schema

Alternatively, do you think it would make sense to not include this in the schema? I'm not sure about when you define a route directly, but if you make an `APIBlueprint`, you should be able to pass in `enable_openapi=False` and it shouldn't be a part of Openapi.

lorenyu (approved):

app/tests/src/route/test_index_route.py
```diff
@@ -0,0 +1,3 @@
+def test_get_index_200(client):
+    response = client.get("/")
+    assert response.status_code == 200
```
chouinar, line 3: Could you add a quick check that the HTML is right? At least that the response body has some of the text?

lorenyu:
> Alternatively, do you think it would make sense to not include this in the schema?

i support excluding this endpoint from the schema if it's easy to do. i really don't care what this endpoint returns, just that it's a 200, since the infra e2e tests assume that the index page returns _something_

PR #180, opened 2023-07-12
Update getting-started.md

daphnegold:
## Ticket

N/A

## Changes
1. Looks like convention now is to use `local.env` and the description of `make setup-local` was outdated
2. remove "Setup" step because that's handled in the running the app steps

## Context
I just ran through the entire setup fresh and this reflects my experience.


jayreed-nava (approved):
LGTM

PR #182, opened 2023-08-08
Increase Postgres connection timeout to 10 seconds

lorenyu:
## Changes
- Increase Postgres connection timeout to 10 seconds
- Remove QueuePool timeout

## Context
Database migrations are occasionally timing out in the platform-test-flask repo.
Tuning this up to 10 seconds to see if that helps.
Also removing the QueuePool timeout which doesn't seem needed.

Example failing migration run:
https://github.com/navapbc/platform-test-flask/actions/runs/5798505305

database migration logs that show timeout expired
image

See 🔒 [slack thread](https://nava.slack.com/archives/C03G1SWD9H7/p1690906746658009)

## Testing

relying on CI since small change


Nava-JoshLong (approved):
LGTM!

PR #185, opened 2023-08-09
Run flask using production wsgi/http server

### LINKED ISSUE #174: Run flask using production wsgi/http server
Flask's built in web server is meant for development, not production (see https://flask.palletsprojects.com/en/2.3.x/deploying/) so we need to install an additional library to make it work in production. A few past projects used [gunicorn](https://flask.palletsprojects.com/en/2.3.x/deploying/gunicorn/) so let's go with that

There are some notes on configuration and design in this 🔒 [slack thread](https://nava.slack.com/archives/C3TRTAJ21/p1689964983526389). tldr; we may want to consider `len(os.sched_getaffinity(0)) * 2` for number of workers, hardcode number of threads (e.g. `4`), and have some sort of override/flag for the command in docker-compose to bypass gunicorn when running locally.

daphnegold:
## Ticket

Resolves #174 

## Changes
- Add and configure gunicorn
- Local and prod agnostic entrypoint in `create_app()`, remove `app_config.environment`
- Remove `__main__.py` (this just adds another layer of complication that does nothing at this point except being an app factory wrapping an app factory)
- Dockerfile starts gunicorn by default
- docker-compose starts hot reloading Flask server, overriding gunicorn cmd in Dockerfile

### Local via docker compose
Screenshot 2023-08-09 at 12 47 29 PM

### Local with Gunicorn
Screenshot 2023-08-09 at 11 23 48 AM

### Gunicorn running on AWS, 1 worker
Screenshot 2023-08-09 at 3 00 47 PM

### Gunicorn running on AWS, 4 workers (available CPU x 2), 4 threads
Screenshot 2023-08-10 at 10 21 57 AM


app/gunicorn.conf.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+from src.app_config import AppConfig
+
+app_config = AppConfig()
+
+bind = app_config.host + ':' + str(app_config.port)
+workers = len(os.sched_getaffinity(0)) * 2
```
lorenyu: can you add comment/references for this value

app/gunicorn.conf.py
```diff
@@ -0,0 +1,8 @@
+import os
```
lorenyu, line 12: can we add a file level docstring about what this file is for and maybe link to https://docs.gunicorn.org/en/stable/configure.html

app/gunicorn.conf.py
lorenyu, line 1: should we define number of threads too? my intuition is that we can do more than one thread

lorenyu (commented):
awesome. can we test the following scenarios (and screenshots for evidence):

- running locally using default method defined in docker compose
- running in aws environment (also include screenshot of logs)
- running locally using gunicorn

app/gunicorn.conf.py
daphnegold: It's defined in the `gunicorn.conf.py` file

daphnegold: Definitely, forgot to hardcode those.

app/gunicorn.conf.py
daphnegold: Thanks for holding us to a high level of documentation! You really set the bar high here.

app/gunicorn.conf.py
```diff
@@ -0,0 +1,24 @@
+"""
+Configuration file for the Gunicorn server used to run the application in production environments.
+
+Attributes:
+    bind(str): The socket to bind. Formatted as '0.0.0.0:$PORT'.
+    workers(int): The number of worker processes for handling requests.
+    threads(int): The number of worker threads for handling requests.
```
lorenyu:
```suggestion
    threads(int): The number of threads per worker for handling requests.
```

app/gunicorn.conf.py
```diff
@@ -0,0 +1,24 @@
...
+
+app_config = AppConfig()
+
+bind = app_config.host + ':' + str(app_config.port)
+# Calculates the number of usable cores and doubles it. The recommended number
+# of workers per core is two. However, adjusting task resources or number of
+# workers may be necessary if they are timing out and crashing.
+# https://docs.gunicorn.org/en/latest/design.html#how-many-workers
+workers = len(os.sched_getaffinity(0)) * 2
+threads = 4
```
lorenyu: nit: add newline at end of file so that we don't have to see the red circle in github :P 
image

```suggestion
threads = 4

```

docs/app/database/database-access-management.md
```diff
@@ -4,7 +4,7 @@ This document describes the best practices and patterns for how the application
...
 
-The database client is initialized when the application starts (see [src/\_\_main\_\_.py](../../../app/src/__main__.py). As the database engine that is used to create acquire connections to the database is initialized using the database configuration defined in [db_config.py](../../../app/src/db/db_config.py), which is configured through environment variables. The initialized database client is then stored on the Flask app's [\`extensions\` dictionary](https://flask.palletsprojects.com/en/2.2.x/src/#flask.Flask.extensions) to be used throughout the lifetime of the application.
+The database client is initialized when the application starts (see [src/\_\_main\_\_.py](../../../app/src/app.py). As the database engine that is used to create acquire connections to the database is initialized using the database configuration defined in [db_config.py](../../../app/src/db/db_config.py), which is configured through environment variables. The initialized database client is then stored on the Flask app's [\`extensions\` dictionary](https://flask.palletsprojects.com/en/2.2.x/src/#flask.Flask.extensions) to be used throughout the lifetime of the application.
```
lorenyu:
```suggestion
The database client is initialized when the application starts (see [src/\_\_main\_\_.py](../../../app/src/app.py). The database engine that is used to create acquire connections to the database is initialized using the database configuration defined in [db_config.py](../../../app/src/db/db_config.py), which is configured through environment variables. The initialized database client is then stored on the Flask app's [\`extensions\` dictionary](https://flask.palletsprojects.com/en/2.2.x/src/#flask.Flask.extensions) to be used throughout the lifetime of the application.
```

lorenyu (approved):
Amazing PR, thanks for the excellent work!

app/gunicorn.conf.py
```diff
@@ -0,0 +1,24 @@
...
+
+from src.app_config import AppConfig
+
+app_config = AppConfig()
+
+bind = app_config.host + ':' + str(app_config.port)
+# Calculates the number of usable cores and doubles it. The recommended number
+# of workers per core is two. However, adjusting task resources or number of
+# workers may be necessary if they are timing out and crashing.
+# https://docs.gunicorn.org/en/latest/design.html#how-many-workers
```
lorenyu, line 20: Oh actually can you add more to the comment to explain why we use os.sched_getaffinity(0) instead of cpu_count()

PR #188, opened 2023-08-11
Set default table privileges in first migration

### LINKED ISSUE #186: Fix table permissions by adding initial migration that sets default permissions
We should add an initial migration that runs "ALERT DEFAULT PRIVILEGES GRANT ALL ON TABLES TO app"

See this PR for context: https://github.com/navapbc/template-infra/pull/377

lorenyu:
## Ticket

Resolves #186 

## Changes

* Add first migration that alters default table privileges for migrator user to grant table privileges to app user
* Combine two subsequent migrations into one file to simplify the template
* Make CI checks run in parallel to improve visibility into what actually failed
* Use docker compose v2

## Context for reviewers

These migration changes don't follow the append-only practice of writing migrations, but since this is a template rather than an active project, I think that makes sense so that we can keep the project's initial setup clean.

## Testing

Ran `make db-recreate` then `make db-upgrade` and `make db-downgrade-all` manually locally

Also this should be covered by CI since we have a test-migrations test


app/src/db/migrations/versions/2023_08_10_default_table_privileges.py
```diff
@@ -0,0 +1,24 @@
...
+revision = "3ed861176e3d"
+down_revision = None
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Change default privileges for future tables created by the `migrator`
+    # user to automatically be accessible by the `app` user.
+    op.execute("ALTER DEFAULT PRIVILEGES GRANT ALL ON TABLES TO app")
```
chouinar, line 25: Do we want to add more rules than just tables?

When I set this up for local development on my project, I did this for each schema, borrowing some of it from our PFML project:
- `REVOKE ALL ON SCHEMA PUBLIC FROM PUBLIC"`
- `GRANT USAGE ON SCHEMA {schema_name} TO {role};` (I think this is automatic for the public schema)
- `ALTER DEFAULT PRIVILEGES IN SCHEMA {schema_name} GRANT ALL ON TABLES TO {role};`
- `ALTER DEFAULT PRIVILEGES IN SCHEMA {schema_name} GRANT ALL ON SEQUENCES TO {role};`
- `ALTER DEFAULT PRIVILEGES IN SCHEMA {schema_name} GRANT ALL ON ROUTINES TO {role};`

Sequences and routines are a bit less necessary, but threw them in

lorenyu: Yeah those are all great. The idea I had was to do as much as possible in the infra template, so the first two items (1) `REVOKE ALL ON SCHEMA public FROM public` and (2) `GRANT USAGE ON SCHEMA {schema_name} TO {role}` are going to be done in template-infra. (2) is already done and (1) is captured by [this ticket](https://github.com/navapbc/template-infra/issues/300).

I'm fine adding SEQUENCES and ROUTINES to this PR.

lorenyu: Ok added

chouinar: Hmm - just thinking of a developer in the future looking at this (or at the piece in the infra), I'd wonder why those two are split into two separate places. Could you add a comment here explaining that? And maybe add to that other ticket to update the comment to link to that bit of code so a developer can see the full picture?

lorenyu: Good point, added a comment and updated the ticket.

lorenyu: @chouinar no rush but lemme know if you're good with the latest changes

chouinar (approved):
LGTM!

PR #191, opened 2023-08-11
Remove Template from CI name

lorenyu:
## Ticket

n/a

## Changes

see title

## Context for reviewers

Remove the word "Template" from the CI name since this is the actual CI that will run on a project.
Also, Flask template doesn't require that the application is an API, so rename to "app".
Make it consistent with [Nextjs template CI name](https://github.com/navapbc/template-application-nextjs/blob/main/.github/workflows/ci-app.yml#L1)

## Testing

n/a

### DIFF
--- a/.github/workflows/ci-app.yml
+++ b/.github/workflows/ci-app.yml
@@ -6,11 +6,11 @@ on:
       - main
     paths:
       - app/**
-      - .github/workflows/api-checks.yml
+      - .github/workflows/ci-app.yml
   pull_request:
     paths:
       - app/**
-      - .github/workflows/api-checks.yml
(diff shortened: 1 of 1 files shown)


sawyerh (approved):

.github/workflows/ci-app.yml
```diff
@@ -1,4 +1,4 @@
-name: Template API Checks
+name: CI - App
 
 on:
   push:
```
sawyerh, line 4: Are the `paths` below correct? I'm wondering if they didn't get updated if/when this filename was changed

lorenyu: good catch will fix

PR #194, opened 2023-09-13
Add a script that waits for the DB container to be available

### LINKED ISSUE #195: Modify how we wait for the Postgres DB locally to check the DB rather wait an arbitrary amount of time

chouinar:
## Ticket

Resolves #195

## Changes

Added a shell script that uses the `pg_isready` command to check if the DB is available rather than arbitrarily waiting 5 seconds.

Added some details for how to run with PY_RUN_APPROACH=local with some local.env changes that occurred a while back.

## Context for reviewers

The notes regarding the PY_RUN_APPROACH and loading environment variables was needed to unblock my development, but only amounts to some updated documentation so should be fine to combine with this PR.

## Testing

Was able to run this locally:
Screenshot 2023-09-13 at 2 55 49 PM

### DIFF
--- a/app/Makefile
+++ b/app/Makefile
@@ -122,7 +122,7 @@ start-db:
 
 # Calls pg_isready to check if the DB is ready to accept connections
 wait-for-db:
-	./wait-for-local-db.sh
+	./bin/wait-for-local-db.sh
 
 ## Destroy current DB, setup new one
 db-recreate: clean-volumes init-db
--- a/app/Makefile
+++ b/app/Makefile
@@ -114,12 +114,15 @@ check: format-check lint test
 #########################
 
 # Docker starts the image for the DB but it's not quite
-# fully ready, so add a 5 second sleep so upgrade doesn't
-# fail because the DB hasn't started yet.
-init-db: start-db sleep-5 db-migrate
+# ready to accept connections so we add a brief wait script
+init-db: start-db wait-for-db db-migrate
 
 start-db:
 	docker-compose up --detach main-db
+
+# Calls pg_isready to check if the DB is ready to accept connections
+wait-for-db:
+	./wait-for-local-db.sh
 
 ## Destroy current DB, setup new one
 db-recreate: clean-volumes init-db
--- a/app/Makefile
+++ b/app/Makefile
@@ -115,13 +115,10 @@ check: format-check lint test
 
 # Docker starts the image for the DB but it's not quite
 # ready to accept connections so we add a brief wait script
-init-db: start-db wait-for-db db-migrate
+init-db: start-db db-migrate
 
 start-db:
 	docker-compose up --detach main-db
-
-# Calls pg_isready to check if the DB is ready to accept connections
-wait-for-db:
 	./bin/wait-for-local-db.sh
 
 ## Destroy current DB, setup new one
--- a/app/wait-for-local-db.sh
+++ b/app/wait-for-local-db.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+# wait-for-local-db
+
+set -e
+
+# Color formatting
+RED='\033[0;31m'
+NO_COLOR='\033[0m'
+
+MAX_WAIT_TIME=30 # seconds
+wait_time=0
+
+# Use pg_isready to wait for the DB to be ready to accept connections
+# We check every 3 seconds and consider it failed if it gets to 30+
+# https://www.postgresql.org/docs/current/app-pg-isready.html
+while ! pg_isready -h localhost -d main-db -q;
+do
+    echo "waiting on Postgres DB to initialize..."
+    sleep 3
+
+    wait_time=$(($wait_time+3))
+    if [ $wait_time -gt $MAX_WAIT_TIME ]
+    then
+        echo -e "${RED}ERROR: Database appears to not be starting up, running \"docker logs main-db\" to see what the issue is${NO_COLOR}"
+        docker logs main-db
+        exit 1
+    fi
+done
+
+echo "Postgres DB is ready after ~${wait_time} seconds"
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 
 # IDE-specific files
 .vscode/*
+.idea
--- a/app/bin/wait-for-local-db.sh
+++ b/app/bin/wait-for-local-db.sh
@@ -8,7 +8,7 @@ RED='\033[0;31m'
 NO_COLOR='\033[0m'
 
 MAX_WAIT_TIME=30 # seconds
-wait_time=0
+WAIT_TIME=0
 
 # Use pg_isready to wait for the DB to be ready to accept connections
 # We check every 3 seconds and consider it failed if it gets to 30+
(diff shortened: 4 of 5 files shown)


app/Makefile
```diff
@@ -114,13 +114,16 @@ check: format-check lint test
...
 
 # Docker starts the image for the DB but it's not quite
-# fully ready, so add a 5 second sleep so upgrade doesn't
-# fail because the DB hasn't started yet.
-init-db: start-db sleep-5 db-migrate
+# ready to accept connections so we add a brief wait script
+init-db: start-db wait-for-db db-migrate
 
 start-db:
 	docker-compose up --detach main-db
```
lorenyu: Now that I look at this with fresh eyes, is there any reason not to put the `wait-for-local-db.sh` call as part of `start-db`? It seems potentially safest that _any_ time we call start-db that we wait for the db to finish booting up before proceeding. Something like


```suggestion
# ready to accept connections so we add a brief wait script
init-db: start-db db-migrate

start-db:
	docker-compose up --detach main-db
	./wait-for-local-db.sh
```

app/wait-for-local-db.sh
lorenyu, line 1: i haven't thought too much about where to put app-specific scripts. do you think root level is best for now, or should it be something like `/app/bin/` or potentially even `/bin/app/`?

app/wait-for-local-db.sh
```diff
@@ -0,0 +1,30 @@
...
+RED='\033[0;31m'
+NO_COLOR='\033[0m'
+
+MAX_WAIT_TIME=30 # seconds
+wait_time=0
+
+# Use pg_isready to wait for the DB to be ready to accept connections
+# We check every 3 seconds and consider it failed if it gets to 30+
+# https://www.postgresql.org/docs/current/app-pg-isready.html
+while ! pg_isready -h localhost -d main-db -q;
```
lorenyu: optional suggestion: what do you think about using the `until` command the way VA uses it (https://github.com/department-of-veterans-affairs/abd-vro/blob/e88312e8d35c71013dcd8748b123f2c77f9de59b/helm/vro-app/templates/deployment.yaml#L38-L41)

i hadn't known about that command before but it reads pretty nicely. i don't feel strongly either way though

app/wait-for-local-db.sh
```diff
@@ -0,0 +1,30 @@
...
+# wait-for-local-db
+
+set -e
+
+# Color formatting
+RED='\033[0;31m'
+NO_COLOR='\033[0m'
+
+MAX_WAIT_TIME=30 # seconds
+wait_time=0
```
lorenyu: nit: i have a preference to have all bash variables be upper case, so this would be `WAIT_TIME`

app/wait-for-local-db.sh
```diff
@@ -0,0 +1,30 @@
...
+# https://www.postgresql.org/docs/current/app-pg-isready.html
+while ! pg_isready -h localhost -d main-db -q;
+do
+    echo "waiting on Postgres DB to initialize..."
+    sleep 3
+
+    wait_time=$(($wait_time+3))
+    if [ $wait_time -gt $MAX_WAIT_TIME ]
+    then
+        echo -e "${RED}ERROR: Database appears to not be starting up, running \"docker logs main-db\" to see what the issue is${NO_COLOR}"
```
lorenyu: grammar nit + small wording suggestion
```suggestion
        echo -e "${RED}ERROR: Database appears to not be starting up. Running \"docker logs main-db\" to troubleshoot.${NO_COLOR}"
```

app/wait-for-local-db.sh
lorenyu, line 1: style nit: i don't have a strong preference for 2 spaces vs 4 spaces, but since [all the scripts in template-infra](https://github.com/navapbc/template-infra/blob/main/bin/) are using 2-space tabs, could we use 2-spaces here for consistency?

lorenyu (approved):
a few tiny comments but looks good thanks!

app/wait-for-local-db.sh
chouinar: I like that, will switch.

app/wait-for-local-db.sh
chouinar: I like troubleshoot, but it runs the command for you on the next line so telling you to try something it does for you might be confusing

app/wait-for-local-db.sh
chouinar: Fine with me

app/Makefile
chouinar: I personally like having each command do just one thing, can help with testing/debugging it. For example, when setting this particular script up, I ran the individual components one-by-one to verify the behavior worked even if it was run repeatedly on an already running DB / no DB exists.

lorenyu: In that case, what about making `wait-for-db` a dependency of `start-db` e.g.

```
start-db: wait-for-db
  ...
```

app/wait-for-local-db.sh
chouinar: I like just putting them in /app/bin/ - clear that they're for the app

app/wait-for-local-db.sh
chouinar: It seems like there are a few conventions, some that prefer lowercase for script-local variables: https://google.github.io/styleguide/shellguide.html#s7-naming-conventions

But I also see a lot of arguments both ways online.

lorenyu: ah ok, but maybe for consistency then we can do all uppercase? (since i didn't realize about the lower case convention when working on infra template bin scripts)

app/Makefile
chouinar: That would still not let you run the commands individually via the Make command. `init-db` is the one that groups all of the more specific ones together as it is organized right now. We could make an intermediate command and have it like:

```Make
init-db: start-and-wait-for-db db-migrate

# come up with a better name
start-and-wait-for-db: start-db wait-for-db

start-db:
	docker-compose up --detach main-db

wait-for-db:
	./bin/wait-for-local-db.sh

```
But if we add too many it might get difficult to follow.

lorenyu: I'm not convinced it's that important to be able to run everything individually. If really needed, it should be possible to just run the shell command directly like `./bin/wait-for-local-db.sh`

If we just kept it simple as:

```
init-db: start-db db-migrate

start-db:
  docker-compose up --detach main-db
  ./bin/wait-for-local-db.sh
```

common scenarios that devs will use on a regular basis should have make targets:

- get everything set up: `make init-db`
- starting the db but not running migrations: `make start-db` (should wait for db to finish booting, which is the common case that devs care about)
- just running migrations: `make db-migrate`

uncommon scenarios that are only relevant when making changes to the make targets themselves don't need special make targets since we can test out the commands manually:

- starting without waiting: `docker-compose up --detach main-db`
- just waiting: `./bin/wait-for-local-db.sh`

app/wait-for-local-db.sh
lorenyu: oh i totally missed that. edited my suggestion above.

lorenyu: nvm you already updated, i'll resolve the comment

app/Makefile
chouinar: That makes sense, switched it to what you're suggesting.

PR #197, opened 2023-09-20
Upgrade to SQLAlchemy 2.0

chouinar:
## Ticket

https://github.com/navapbc/template-application-flask/issues/82

## Changes

Upgraded to SQLAlchemy 2.0, see context section for major changes in this version

## Context for reviewers

SQLAlchemy 2.0 comes with an incredibly thorough migration guide: https://docs.sqlalchemy.org/en/20/changelog/migration_20.html

Thankfully 1.4 started nudging usage of SQLAlchemy to use the new approaches, so the changes needed weren't that large.

Noteworthy changes:
* DB model definitions have been adjusted. The python types now need to be wrapped in `Mapped[...]`, and columns now get defined either with `mapped_column` OR nothing with SQLAlchemy figuring out the type from the python type (similar to Dataclasses or Pydantic). This largely results in more minimal class definitions especially ones with a bunch of basic columns.
* DB session/connection/engine logic adjusted slightly to have fewer edge cases. For the changes here that just required some adjustments to the underlying connection setup, something I'd done on another project already to fix some connection issues. Largely this comes down to connections no longer auto-committing and needing to be in `with conn.begin()` blocks to function properly.
* Raw SQL calls to `execute` should be wrapped in the `text` class
* DeclarativeBase is better supported and instantiated slightly differently. The Metadata object no longer works globally and has to be attached to the DeclarativeBase to work properly.
* Typing for MyPy is built into SQLAlchemy and no longer relies on libraries like SQLAlchemy-stubs

## Testing

Since the DB hits virtually everything, I tested everything as thoroughly as possible.

Because of how the SQLAlchemy-stubs we removed work, they need to be completely deleted otherwise MyPy thinks they should run. If you are running outside of the docker container, run `poetry install --no-root --all-extras --with dev --sync` which will delete extra packages.

### Basics
Tests, formatting, linting, all working, only required a few fixes to make SQLAlchemy happy.

### SQLAlchemy warnings
When running unit tests, SQLAlchemy will output warnings for deprecated features. The only still usable deprecated feature we had was our `get`queries which were adjusted. Prior to the fix, we would see this warning:
Screenshot 2023-09-20 at 11 00 38 AM

### Migrations
They still work, adding a few new columns generates what we would expect and uses the mapping config added to the Base class.
Screenshot 2023-09-19 at 3 36 16 PM

Screenshot 2023-09-19 at 3 36 11 PM

### Swagger
Was able to successfully use the local swagger endpoints and create/update/read from the DB which was populated like so:
Screenshot 2023-09-20 at 11 34 39 AM


app/src/db/migrations/run.py
```diff
@@ -53,41 +50,3 @@ def have_all_migrations_run(db_engine: sqlalchemy.engine.Engine) -> None:
         logger.info(
             f"The current migration head is up to date, {current_heads} and Alembic is expecting {expected_heads}"
         )
-
-
-def check_model_parity() -> None:
```
chouinar, line 58: This isn't used and was giving me some type errors. There's a better way to do this anyways as Alembic has a utility now for checking the model parity for you which I'll create a ticket and add later.

app/src/adapters/db/clients/postgres_client.py
```diff
@@ -46,9 +46,6 @@ def get_conn() -> Any:
         return sqlalchemy.create_engine(
             "postgresql://",
             pool=conn_pool,
-            # FYI, execute many mode handles how SQLAlchemy handles doing a bunch of inserts/updates/deletes at once
-            # https://docs.sqlalchemy.org/en/14/dialects/postgresql.html#psycopg2-fast-execution-helpers
-            executemany_mode="batch",
```
chouinar, line 51: Batch mode no longer exists: https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#psycopg2-fast-execution-helpers

There's another way to configure it to do a similar operation, but I wasn't sure if we wanted to keep it

app/tests/lib/db_testing.py
```diff
@@ -25,21 +27,28 @@ def create_isolated_db(monkeypatch) -> db.DBClient:
...
         _create_schema(conn, schema_name)
-        try:
-            yield db_client
-        finally:
+
+    try:
+        yield db_client
+
+    finally:
+        with db_client.get_connection() as conn:
```
lorenyu: intuitively seems dangerous to try to connect to the db in a finally clause. did you encounter an issue where we need a new connection in the cleanup?

app/src/db/models/base.py
```diff
@@ -26,10 +25,31 @@
...
+    # See: https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#mapped-column-derives-the-datatype-and-nullability-from-the-mapped-annotation
+    #      for the default mappings
+    #
+    # See: https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#orm-declarative-mapped-column-type-map
+    #      for details on setting up this configuration.
+    type_annotation_map = {
+        # Explicitly use the Text column type for strings
+        str: Text,
+        # Always include a timezone for datetimes
+        datetime: TIMESTAMP(timezone=True),
```
lorenyu, line 46: nonblocking question: any idea why sqlalchemy makes `TIMESTAMP` all caps?

app/src/db/models/base.py
```diff
@@ -26,10 +25,31 @@
...
+    #   my_column: Mapped[str]
+    #
+    # If you provide a mapped_column attribute you can override these values
+    #
+    # See: https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#mapped-column-derives-the-datatype-and-nullability-from-the-mapped-annotation
+    #      for the default mappings
+    #
+    # See: https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#orm-declarative-mapped-column-type-map
+    #      for details on setting up this configuration.
+    type_annotation_map = {
```
lorenyu, line 44: should we map uuids too?

```
uuid.UUID: postgresql.UUID
```

then we can simplify the model declarations by not needing to specify the type, e.g.

```
 id: Mapped[uuid.UUID] = mapped_column(
        primary_key=True, default=uuid.uuid4
    )
```

and 

```
user_id: Mapped[UUID] = mapped_column(
        ForeignKey("user.id", ondelete="CASCADE"), primary_key=True
    )
```

app/src/db/models/base.py
```diff
@@ -72,17 +94,17 @@ def same_as_created_at(context: Any) -> Any:
 @declarative_mixin
 class TimestampMixin:
     """Mixin to add created_at and updated_at columns to a model
-    https://docs.sqlalchemy.org/en/14/orm/declarative_mixins.html#mixing-in-columns
+    https://docs.sqlalchemy.org/en/20/orm/declarative_mixins.html#mixing-in-columns
     """
 
-    created_at: datetime = Column(
+    created_at: Mapped[datetime] = mapped_column(
         TIMESTAMP(timezone=True),
```
lorenyu: i think we can remove this now that we defaulted datetime to map to TIMESTAMP

app/src/db/models/base.py
```diff
@@ -72,17 +94,17 @@ def same_as_created_at(context: Any) -> Any:
...
         nullable=False,
         default=datetime_util.utcnow,
         server_default=sqlnow(),
     )
 
-    updated_at: datetime = Column(
+    updated_at: Mapped[datetime] = mapped_column(
         TIMESTAMP(timezone=True),
```
lorenyu: i think we can remove this now that we defaulted datetime to map to TIMESTAMP

app/src/db/models/base.py
```diff
@@ -59,10 +79,12 @@ def copy(self, **kwargs: dict[str, Any]) -> "Base":
 @declarative_mixin
 class IdMixin:
     """Mixin to add a UUID id primary key column to a model
-    https://docs.sqlalchemy.org/en/14/orm/declarative_mixins.html
+    https://docs.sqlalchemy.org/en/20/orm/declarative_mixins.html
     """
 
-    id: uuid.UUID = Column(postgresql.UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    id: Mapped[uuid.UUID] = mapped_column(
+        postgresql.UUID(as_uuid=True), primary_key=True, default=uuid.uuid4
```
lorenyu: see comment above about adding uuid to the default list of mapped types, in which case we can just do
```suggestion
        primary_key=True, default=uuid.uuid4
```

app/src/db/models/user_models.py
```diff
@@ -21,21 +21,21 @@ class RoleType(str, enum.Enum):
...
+    roles: Mapped[list["Role"]] = relationship(
         "Role", back_populates="user", cascade="all, delete", order_by="Role.type"
     )
 
 
 class Role(Base, TimestampMixin):
     __tablename__ = "role"
-    user_id: Mapped[UUID] = Column(
+    user_id: Mapped[UUID] = mapped_column(
         postgresql.UUID(as_uuid=True), ForeignKey("user.id", ondelete="CASCADE"), primary_key=True
```
lorenyu: see comment above about adding uuid to the default list of mapped types, in which case we can just do
```suggestion
       ForeignKey("user.id", ondelete="CASCADE"), primary_key=True
```

app/tests/conftest.py
```diff
@@ -80,7 +80,7 @@ def db_client(monkeypatch_session) -> db.DBClient:
     """
 
     with db_testing.create_isolated_db(monkeypatch_session) as db_client:
-        models.metadata.create_all(bind=db_client.get_connection())
+        models.metadata.create_all(bind=db_client._engine)
```
lorenyu: can we keep this to passing the connection so that we don't have to expose the the private property of the client? i'm reluctant to change the interface of DBClient to make the engine public just for the test suite – currently engine is private (hence the underscore prefix).

lorenyu (approved):
One request to not access the private member `_engine` of `DBClient`, a question about establishing in a new connection in a finally clause, and a few other small suggestions, but otherwise LGTM!

Note for when we generate release notes, we might want to point out that this is a breaking change and provide resources for how to migrate (e.g. link to SQLAlchemy migration docs), similar to how we did it on template-infra (see 
https://github.com/navapbc/template-infra/releases/tag/v0.3.0)

app/tests/lib/db_testing.py
chouinar: I agree, but I'm not sure if there's any alternative. We need a connection to run the delete schema command. If this were to fail, we'd just leave that schema behind. Due to me missing something for the delete schema, while I was testing this I ended up with 50 test schemas and it didn't cause any issues, so worst case is not cleaning up properly

app/src/db/models/base.py
chouinar: I'm not sure, in some cases they have a normal-cased one like with Text, but not in this case.

app/tests/conftest.py
chouinar: What was there before doesn't work anymore, I'll see if there is another way to use the connection more directly.

chouinar: Looks like there is a way
```
        with db_client.get_connection() as conn, conn.begin():
            models.metadata.create_all(bind=conn)
```

app/src/db/models/base.py
chouinar: Yknow, I didn't even interpret the mapping as working like that and auto-populating the first field for mapped_column. Good catch, fixed in each of the spots you mentioned.

app/tests/lib/db_testing.py
lorenyu: > 'm not sure if there's any alternative

did the original code not work? in the old code, we indented the try/catch block so that we can use the same connection that created the schema to drop the schema e.g.:

```python
    with db_client.get_connection() as conn:
        _create_schema(conn, schema_name)

        try:
            yield db_client

        finally:
            _drop_schema(conn, schema_name)
```

i guess i'm trying to understand why we need to exit the connection context before we do the try/catch

app/tests/conftest.py
lorenyu: awesome

app/src/db/models/base.py
```diff
@@ -26,10 +25,33 @@
...
+    # If you provide a mapped_column attribute you can override these values
+    #
+    # See: https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#mapped-column-derives-the-datatype-and-nullability-from-the-mapped-annotation
+    #      for the default mappings
+    #
+    # See: https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#orm-declarative-mapped-column-type-map
+    #      for details on setting up this configuration.
+    type_annotation_map = {
+        # Explicitly use the Text column type for strings
+        str: Text,
```
lorenyu: optional style nit: should we alphabetize this map e.g.

```
{
  datetime: ...
  str: ...
  uuid.UUID: ...
}
```

no big deal either way the map is pretty small now, but it might be nice if it ever gets big

app/tests/lib/db_testing.py
chouinar: I had originally done that in the NJ codebase and ran into errors if the connection failed, but trying again (running a unit test, pause in the middle, deleting a table from the schema so the client fails) still dropped the schema. So either something is slightly different here, or 2.0 fixed whatever problem I was hitting. Either way, I can do that.

lorenyu (approved):
Looks great thanks!

PR #201, opened 2023-09-21
Upgrade to psycopg3 from version 2

chouinar:
## Ticket

https://github.com/navapbc/template-application-flask/issues/200

## Changes

Upgraded psycopg, the library that connects to Postgres for SQLAlchemy, to version 3.

## Context for reviewers

Version 2 was named `psycopg2` but version 3 is just `psycopg` so that they can upgrade without changing the package name moving forward.

Changing the connection string to `postgresql+psycopg` tells SQLAlchemy internally to use the new library: https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#dialect-postgresql-psycopg-connect

Since we only have a few direct interactions with psycopg, the changes were fairly minimal, just adjusting the DB connection info object slightly.

## Testing

Migrations, unit tests, and other automations still running fine.

Was able to pull up swagger and read/write to the DB via our endpoints.

### DIFF
--- a/app/poetry.lock
+++ b/app/poetry.lock
@@ -1070,71 +1070,90 @@ dev = ["pre-commit", "tox"]
 testing = ["pytest", "pytest-benchmark"]
 
 [[package]]
-name = "psycopg2-binary"
-version = "2.9.3"
-description = "psycopg2 - Python-PostgreSQL Database Adapter"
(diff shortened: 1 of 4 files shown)


lorenyu (approved):
Nice

PR #202, opened 2023-09-22
Fix verify_in_ssl log statement which no longer has access to SSL params

chouinar:
## Ticket

https://github.com/navapbc/template-application-flask/issues/200

## Changes

Removes log statement that relies on variables no longer in psycopg 3.

## Context for reviewers
The upgrade to psycopg3 got this error during migrations: `AttributeError: 'ConnectionInfo' object has no attribute 'ssl_attribute_names'` - https://github.com/navapbc/platform-test-flask/actions/runs/6275581568/job/17043454952

These aren't available anymore and looking through the connection info object, I don't see any equivalent. Looking at the docs for version 2, these attributes say `Only available if psycopg was built with libpq >= 9.5` https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.ConnectionInfo.ssl_attribute 

While the docs for psycopg3 mention libpq as being the backing library it uses, I don't see any mention of these parameters in their docs. https://www.psycopg.org/psycopg3/docs/api/connections.html#psycopg.Connection.pgconn - so seems to just be gone entirely?

## Testing
Updated tests

### DIFF
--- a/app/src/adapters/db/clients/postgres_client.py
+++ b/app/src/adapters/db/clients/postgres_client.py
@@ -121,12 +121,6 @@ def generate_iam_auth_token(aws_region: str, host: str, port: int, user: str) ->
 def verify_ssl(connection_info: Any) -> None:
     """Verify that the database connection is encrypted and log a warning if not."""
     if connection_info.pgconn.ssl_in_use:
-        logger.info(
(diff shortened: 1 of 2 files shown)


lorenyu (approved):
Nice thanks for the fix

PR #206, opened 2023-10-10
Make updates for Python 3.12 & update various packages to their latest versions

chouinar:
## Ticket

https://github.com/navapbc/template-application-flask/issues/205

## Changes

Make package updates to make Python 3.12 work without warnings

## Context for reviewers

Package updates were done via `poetry add <package>@latest [--group dev]` (the group being included for dev dependencies). I did skim the changelogs for any of the libraries we updated, at least the major versions to see if there was anything signficantly breaking, but most of the packages are pretty minimal in changes (bug fixes, small feature updates).

The package updates fall into a few categories:
1. Broken due to new version of Python (cffi was the main one)
2. Warnings through linting or pytest deprecation warnings - newer package versions had corrected the issue (SQLAlchemy, APIFlask, pytz, pytest, mypy)
3. Package only had a few minor updates, figured we might as well update (alembic, isort, flake8, Faker, factory-boy)

The only package of note that I didn't update was Pydantic, which I'll do in a follow-up as I know it has several breaking changes.

This also resolves a few Dependabot alerts for the gitpython, cryptography and urllib3 packages (package.lock has been updated to have the latest version for each of those past the required ones).

## Testing

Running `make format lint test` both inside and outside of the Docker container to verify that there are no issues reported / pytest warnings given. Will verify that this builds as well in CI successfully when the PR is created / later merged.


jamesbursa (approved):
Thanks!

PR #220, opened 2023-12-22
Generate database schema diagrams from SQLAlchemy models

### LINKED ISSUE #219: Generate database schema diagrams from SQLAlchemy models

chouinar:
## Ticket

Resolves #219

## Changes

Generate database schema diagrams (ERDs) from our SQLAlchemy tables

Automate this in our CI/CD

## Context for reviewers

I added this in the simpler grants.gov codebase that is derived from this: https://github.com/HHS/simpler-grants-gov/pull/824 and this logic is originally from a few of our other projects that the template API is based on.

## Testing

The image within this PR was generated by the script which just requires running `make db-create-schema-diagrams`

The workflow ran on this PR: https://github.com/navapbc/template-application-flask/actions/runs/7302558921/job/19901412292


app/pyproject.toml
```diff
@@ -40,6 +40,8 @@ pytest-watch = "^4.2.0"
 pytest-lazy-fixture = "^0.6.3"
 types-pyyaml = "^6.0.12.11"
 setuptools = "^68.2.2"
+sadisplay = "0.4.9"
```
lorenyu, line 43: [looks like this tool is archived](https://github.com/GoodRx/sadisplay)
image

[eralchemy](https://github.com/Alexis-benoist/eralchemy) looks like it has a lot more stars/forks, and the diagrams look better.

Taking a step back, we'll probably need something like this regardless of the tech stack so that we don't have to maintain two separate implementations for python and next.js templates (and in the future we'll probably have a java template too), so I'd like to explore a language-agnostic one that runs in the infra template.

From browsing [this reddit post](https://www.reddit.com/r/PostgreSQL/comments/un877d/generate_erd_from_existing_database_schema_what/), it seems that https://github.com/schemaspy/schemaspy is pretty nice. I like that the generated documentation is also interactive. For example, the diagram we have on MA PFML is pretty impossible to navigate and read since it's so big, so having interactive features could be really nice. Thoughts?

chouinar: Of course it was archived 4 days after I added it to the Grants repo - If I've got time I'll look into Schema spy in the near future

PR #222, opened 2024-04-04
VSCode API Debugger

rylew1:
## Ticket

Resolves #TODO

## Changes

- Add API debugging ability from VSCode

## Do we want to commit IDE specific files (`.vscode/launch.json`) ?
- I don't see it in the frontend template project

## Context for reviewers

- It seems you can't run `run-logs` alongside `start-debug` for some reason - the 5678 port won't be exposed

- Demo running `make start-debug`, then connect with the VSCode launcher:

https://github.com/navapbc/template-application-flask/assets/93001277/c411675c-7e32-42ae-a70c-267bc5954486


app/src/app.py
```diff
@@ -34,6 +40,29 @@ def create_app() -> APIFlask:
     return app
 
 
+def is_running_in_local() -> bool:
+    print('s.getenv("FLASK_ENV") => ', os.getenv("FLASK_ENV"))
+    return os.getenv("FLASK_ENV") in ("local", "dev")
```
chouinar: The environment is just `ENVIRONMENT`. I'd also restrict it to only `local` as some projects use `dev` as a non-local environment.

docker-compose.debug.yml
```diff
@@ -0,0 +1,21 @@
+version: "3.8"
+
+# run with `docker compose -f`
+# combines ports and env vars with main docker-compose.yml mass-pfml-api service
```
chouinar:
```suggestion
# combines ports and env vars with main docker-compose.yml main-app service
```

app/src/app.py
```diff
@@ -21,6 +22,11 @@
 def create_app() -> APIFlask:
     app = APIFlask(__name__)
 
+    print("app => ", app)
+
+    if is_running_in_local():
+        setup_debug_adapter()
```
chouinar: We might want to put this `is_running_in_local` function in `local.py`.

https://github.com/navapbc/template-application-flask/blob/main/app/src/util/local.py

app/Makefile
```diff
@@ -92,6 +92,9 @@ build:
 start:
 	docker compose up --detach
 
+start-debug:
+	docker compose -f ../docker-compose.yml -f ../docker-compose.debug.yml up --detach
+
```
rylew1, line 97: debug file gets launched second to be more clear that it overrides value in the normal docker compose

chouinar (approved):

PR #224, opened 2024-05-08
Add bulk operations utilities

KevinJBoyer:
## Ticket

n/a

## Changes

 - Add `bulk_ops.py`, which exposes a `bulk_upsert` function for efficiently upserting large amounts of data into the database

## Context for reviewers

 - Projects frequently need to read in large amounts of data into the database from external sources such as CSV files. This utility provides a flexible way of doing so efficiently.
 - I'm not familiar with the platform's approach to abstracting away the underlying database -- the code here is Postgres specific and uses the `psycopg` library. Feedback on how to adapt the code here to the platform's approach is welcome/appreciated.

## Testing
`make test args="tests/src/db/test_bulk_ops.py"`
image


app/tests/src/db/test_bulk_ops.py
```diff
@@ -0,0 +1,89 @@
...
+from psycopg import rows, sql
+
+import src.adapters.db as db
+from src.db import bulk_ops
+
+
+def test_bulk_upsert(db_session: db.Session):
+    conn = db_session.connection().connection
+    # Override mypy, because SQLAlchemy's DBAPICursor type doesn't specify the row_factory attribute, or that it functions as a context manager
+    with conn.cursor(row_factory=rows.class_row(Number)) as cur:  # type: ignore
```
KevinJBoyer: Would love to know if there's a better way of doing this. I also considered:
```
    db_client = db.PostgresDBClient()
    conn = db_client._engine.raw_connection()
```
but accessing `_engine` directly did not feel appropriate (and doesn't solve for the type issue in any case)

app/src/db/bulk_ops.py
```diff
@@ -0,0 +1,150 @@
...
+        cur,
+        src_table=temp_table,
+        dest_table=table,
+        columns=attributes,
+        constraint=constraint,
+        update_condition=update_condition,
+    )
+
+
+def create_temp_table(cur: psycopg.Cursor, temp_table: str, src_table: str) -> None:
```
lorenyu: can we prefix non-public methods with an underscore

app/src/db/bulk_ops.py
```diff
@@ -0,0 +1,150 @@
+"""Bulk database operations for performance.
+
+Provides a bulk_upsert function
+"""
```
lorenyu, line 5: maybe update doc for this module or for the bulk_upsert function that this is postgres + psycopg specific

app/src/db/bulk_ops.py
```diff
@@ -0,0 +1,150 @@
...
+from typing import Any, Optional, Sequence
+
+import psycopg
```
lorenyu, line 8: Should we explicitly include psycopg in Poetry dependencies file rather than rely on it being included by SQLAlchemy? 🤔

app/tests/src/db/test_bulk_ops.py
lorenyu: hmm, not sure

app/tests/src/db/test_bulk_ops.py
```diff
@@ -0,0 +1,89 @@
...
+            obj.num = random.randint(1, 10000)
+
+        bulk_ops.bulk_upsert(
+            cur,
+            table,
+            attributes,
+            objects,
+            constraint,
+        )
+        conn.commit()
```
lorenyu: nit: it'd be nice to have the test case do a combination of inserts and updates rather than just inserts and updates separately

app/src/db/bulk_ops.py
```diff
@@ -0,0 +1,150 @@
...
+      objects: a sequence of objects to upsert
+      constraint: the table unique constraint to use to determine conflicts
+      update_condition: optional WHERE clause to limit updates for a
+        conflicting row
+    """
+    if not update_condition:
+        update_condition = sql.SQL("")
+
+    temp_table = f"temp_{table}"
+    create_temp_table(cur, temp_table=temp_table, src_table=table)
```
chouinar: This is probably a very niche edge case, but what would happen if two temp tables were created with the same name by different processes? Does that cause any issues, or does them being in the transactions entirely shield them?

app/src/db/bulk_ops.py
```diff
@@ -0,0 +1,150 @@
...
+kwargs_row = rows.kwargs_row
+
+
+def bulk_upsert(
+    cur: psycopg.Cursor,
+    table: str,
+    attributes: Sequence[str],
+    objects: Sequence[Any],
+    constraint: str,
+    update_condition: Optional[sql.SQL] = None,
```
chouinar: The newer form for optional/nullable fields saves an import.
```suggestion
    update_condition: sql.SQL | None = None,
```

app/tests/src/db/test_bulk_ops.py
chouinar: We could consider adding a `raw_connection()` method to the client class which does what you suggested. For the docs, mention that unless you're trying to do something very low level (ie. in psycopg) you'll almost never actually want to use it.

app/tests/src/db/test_bulk_ops.py
```diff
@@ -0,0 +1,89 @@
...
+class Number:
+    id: str
+    num: int
+
+
+def get_random_number_object() -> Number:
+    return Number(
+        id=str(random.randint(1000000, 9999999)),
+        num=random.randint(1, 10000),
+    )
```
chouinar: Should we put these test utils at the top of the file?

app/src/db/bulk_ops.py
KevinJBoyer: I believe it already is -- https://github.com/navapbc/template-application-flask/pull/201/files#diff-f79644217c104a26d0814d8faf2afc24c0239350806e80ff239b93271888e465R22

app/src/db/bulk_ops.py
KevinJBoyer: TIL, ty!

app/src/db/bulk_ops.py
KevinJBoyer: Great question, I tested it locally and it looks like the transaction isolation works like you'd expect. Here's the SQL I ran:

```BEGIN;
CREATE TEMP TABLE test (id INT) ON COMMIT DROP;
SELECT * FROM test;

-- In a separate connection!
BEGIN;
CREATE TEMP TABLE test (other INT) ON COMMIT DROP;
SELECT * FROM test;
COMMIT;

-- Back in the original connection
COMMIT;

File: app/tests/src/db/test_bulk_ops.py (hunk 5 above)
Comment (KevinJBoyer) [reply app/tests/src/db/test_bulk_ops.py]: Added -- one round of inserts, then a second round of combo insert + updates

File: app/tests/src/db/test_bulk_ops.py (hunk 1 above)
Comment (KevinJBoyer) [reply app/tests/src/db/test_bulk_ops.py]: I added this and included a comment with that context -- LMK what you think!

File: app/tests/src/db/test_bulk_ops.py [hunk 9]
```diff
@@ -0,0 +1,94 @@
...
+
+
+def test_bulk_upsert(db_session: db.Session):
+    db_client = db.PostgresDBClient()
+    conn = db_client.get_raw_connection()
+    # conn = db_session.connection().connection
```
lorenyu: nit: cleanup

app/tests/src/db/test_bulk_ops.py
```diff
@@ -0,0 +1,94 @@
...
+        objects.extend([get_random_number_object() for i in range(50)])
+
+        bulk_ops.bulk_upsert(
+            cur,
+            table,
+            attributes,
+            objects,
+            constraint,
+        )
+        conn.commit()
```
lorenyu: nit: issue i have with this testis that even though you only modify half of the objects, you're still doing an update on all the objects since you're doing upsert with the full object list. i think we should only pass the modifed objects to the upsert call and make sure the old objects are still there.

put another way, i could introduce a bug without breaking this test i.e. if i replaced the upsert implementation with a "truncate then insert" that will still work, since the current test passes in all the old unmodified objects.

to make the test cleaner, i recommend separating out the variables you are using to track the expected state of the db and the variables you are using as input parameters. right now you're using `objects` for both input and expected values. similarly, other than the expected state variable which we expect to change, consider treating all variables as constant / don't modify them. in this vein, consider using namedtuples instead of dataclasses which are immutable.

then some optional nice-to-haves, might be nice to pull random objects rather than just the first half of them using random.sample, and mix the updates and inserts.

here's some pseudocode that addresses these ideas:

```
objects_by_id = {}

objects_to_insert = [get_random_number_object() for i in range(100)]
bulk_upsert(..., objects_to_insert, ...)

for obj in objects_to_insert:
  objects_by_id[obj.id] = obj
expected_records = sorted(objects_by_id.values, key=attrgetter("id"))

assert fetch_all_records() == expected_records

objects_to_update = [obj._replace(num=random.randint(1,10000)) for obj in random.sample(objects_by_id.values(), 50)]
more_objects_to_insert = [get_random_number_object() for i in range(50)]
objects_to_upsert = objects_to_update + more_objects_to_insert
random.shuffle(objects_to_upsert)

bulk_upsert(..., objects_to_upsert, ...)

for obj in objects_to_upsert:
  objects_by_id[obj.id] = obj
expected_records = sorted(objects_by_id.values, key=attrgetter("id"))

assert fetch_all_records() == expected_records
```

lorenyu (approved):
looks great. just a nit on the test, don't feel too strongly about it though

app/tests/src/db/test_bulk_ops.py
KevinJBoyer: 👍 makes sense! I made these changes

PR #226, opened 2024-06-06
Bump Python version in pyproject.toml

KevinJBoyer:
## Ticket

n/a

## Changes

Changes minimum Python version in pyproject.toml to 3.12.
Fixes two minor unrelated typos in documentation.

## Context for reviewers

The repo includes a `.python-version` set to 3.12, so the minimum version Poetry requires of dependencies should match.

Alternatively, we could downgrade the .python-version to 3.10, but that seems unnecessary.

## Testing

All existing tests pass.


chouinar (approved):
LGTM!

PR #227, opened 2024-06-14
Add support for local secrets

KevinJBoyer:
## Ticket

n/a

## Changes
 - Provide an example docker compose override file
 - Update .gitignore to ensure non-example override files aren't committed
 - Update documentation (plus comment in local.env) to explain how to use it
 - Move the docker compose files to inside the app folder
 - Update the developer set up instructions

## Context for reviewers
 - This allows developers to set local secrets that are passed to their Docker container.

## Testing
Follow the instructions in the documentation and verify the contents of your .env match what's available to the Docker container.


.gitignore
```diff
@@ -5,3 +5,8 @@
 # IDE-specific files
 .vscode/*
 .idea
+
+# This file is used in local development to pass an /app/.env
+# file to the container, for secrets. It should not be committed
+# to the repo because tests and CI/CD will not have an .env file.
+docker-compose.override.yml
```
chouinar: So, I wonder if the real issue here is where we keep the docker-compose file. Would there be any issue moving it to the app folder itself?

That's what we did on the simpler grants gov repo (this `api` folder is the equivalent of `app`): https://github.com/navapbc/simpler-grants-gov/tree/main/api

We do have a top-level docker-compose but just for the purposes of spinning up both the frontend and API together.

app/local.env
```diff
@@ -63,8 +63,12 @@ HIDE_SQL_PARAMETER_LOGS=TRUE
 # add them to this file to avoid mistakenly
 # committing them. Set these in your shell
 # by doing `export AWS_ACCESS_KEY_ID=whatever`
+# if you are running the app directly, or
+# in your `app/.env` if you are running the
+# app in a Docker container
```
chouinar, line 68: Threading arbitrarily, - we might also need a change to the `load_local_env_vars` method which handles getting the `local.env` file into your env vars when you run outside of docker. Right now that automatically gets called for unit tests in the conftest file:

https://github.com/navapbc/template-application-flask/blob/main/app/tests/conftest.py#L21

I think if we just adjusted `load_local_env_vars` to do "if .env exists, load it before the local.env file" bit - then it would work.

chouinar: Alternatively - we _could_ get rid of load_local_env_vars entirely and more accurately document how to do that properly when running outside of docker. Maybe I'll take that on next week.

docker-compose.override.yml.example
```diff
@@ -0,0 +1,8 @@
+version: '3'
+
+services:
+
+  main-app:
+    env_file:
+      - ./app/local.env
+      - ./app/.env
```
chouinar: I'm curious if we want to name the override env file something like `override.env` to make it clear on the behavior.

.gitignore
lorenyu: i'm open to that

lorenyu (commented):
apologies in advance for not providing more actionable or constructive feedback, but this approach feels kinda hacky and non-standard to me. i don't think this is a typical structure you see for projects using docker and docker compose. i prefer to come up with a "boring" solution that isn't something someone needs to learn if they come from another place.

.gitignore
KevinJBoyer: I agree that inside the app folder seems more appropriate to me.

docs/app/getting-started.md
```diff
@@ -21,17 +19,25 @@ A very simple [docker-compose.yml](/docker-compose.yml) has been included to sup
...
 3. If you are using an M1 mac, you will need to install postgres as well: `brew install postgresql` (The psycopg2-binary is built from source on M1 macs which requires the postgres executable to be present)
 
-4. You'll also need [Docker Desktop](https://www.docker.com/products/docker-desktop/)
+4. You'll also need [Docker Desktop](https://www.docker.com/products/docker-desktop/) installed and running.
 
 ## Run the application
 
-1. In your terminal, `cd` to the `app` directory of this repo.
-2. Make sure you have [Docker Desktop](https://www.docker.com/products/docker-desktop/) installed & running.
-3. Run `make setup-local` to install dependencies
```
KevinJBoyer, line 30: Removed these lines:
 - 1 is redundant with the note about running from within the app folder (which I also moved to be directly above these steps)
 - 2 is redundant with the prerequisites above. If someone does miss that note, the error message is very clear ("Is Docker running?")
 - I *believe* 3 is not necessary since these steps assume running everything in the Docker container. For folks who do want to run things like formatting, etc, directly on their machine, the linked application docs under Next steps at the end of this do a great job explaining how to do so (including doing this step)

docs/app/getting-started.md
```diff
@@ -21,17 +19,25 @@ A very simple [docker-compose.yml](/docker-compose.yml) has been included to sup
...
-4. Run `make init start` to build the image and start the container.
-5. Navigate to `localhost:8080/docs` to access the Swagger UI.
-6. Run `make run-logs` to see the logs of the running API container
-7. Run `make stop` when you are done to delete the container.
+**Note:** Run everything from within the `/app` folder:
+
+1. Run `make init start` to build the image and start the container.
+2. Navigate to `localhost:8080/docs` to access the Swagger UI.
+3. Run `make run-logs` to see the logs of the running API container
+4. Run `make stop` when you are done to delete the container.
```
chouinar: This line is a bit misleading. Deleting might make someone think that the contents of their container will be removed when they run stop, but that isn't true as that's cached and/or stored in the volume.

Maybe we should adjust this to something like `.. when you are done to stop the container.`

chouinar (approved):
LGTM!

docs/app/getting-started.md
KevinJBoyer: 👍  will make this change!

PR #229, opened 2024-06-17
Fix flaky test test_bulk_upsert

KevinJBoyer:
## Ticket

n/a

## Changes
 - Make the IDs deterministic and guaranteed to be unique

## Context for reviewers
 - Since the IDs were random, there was a small chance (but [higher than you might intuitively expect](https://en.wikipedia.org/wiki/Birthday_problem)!) of generating a collision. This would trigger integrity errors or cause the behavior of the test to be not what you'd expect (a new "insert" would actually be a "modify", for example)

## Testing
First, (temporarily) add pytest-repeat:
`poetry add pytest-repeat`
Then run the test several thousand times:
`make test args="tests/src/db/test_bulk_ops.py --count 2500"`
You should observe some failures because a duplicate ID is generated:
Screenshot 2024-06-17 at 8 37 58 AM
Then, re-run after applying the patch:
Screenshot 2024-06-17 at 8 53 47 AM

### DIFF
--- a/app/tests/src/db/test_bulk_ops.py
+++ b/app/tests/src/db/test_bulk_ops.py
@@ -15,9 +15,14 @@ class Number:
     num: int
 
 
+_next_id = 0
+
+
 def get_random_number_object() -> Number:
+    global _next_id
+    _next_id += 1
     return Number(
-        id=str(random.randint(1000000, 9999999)),
+        id=str(_next_id),
         num=random.randint(1, 10000),
     )
 


app/tests/src/db/test_bulk_ops.py
```diff
@@ -15,9 +15,14 @@ class Number:
...
 
+_next_id = 0
+
+
 def get_random_number_object() -> Number:
+    global _next_id
+    _next_id += 1
     return Number(
-        id=str(random.randint(1000000, 9999999)),
+        id=str(_next_id),
```
chouinar, line 25: (not required - just throwing an idea out) You could use Factory boy for building these. While we normally use it for DB models, it also supports creating a dict. Mostly would help with not needing to manage a global counter.
https://github.com/navapbc/simpler-grants-gov/blob/main/api/tests/src/db/models/factories.py#L1269

Not really necessary for just this, but if this was expanded significantly in use I'd recommend it.

chouinar (approved):
LGTM!

app/tests/src/db/test_bulk_ops.py
KevinJBoyer: Oh nice! I'll defer that change for now, thanks for sharing (also, very convenient that the grants.gov project is open source)

PR #232, opened 2024-06-20
Swap out Flake8 for ruff in linting

### LINKED ISSUE #107: Consider switching from flake8 to ruff
apparently there's a new much faster linter for python: https://github.com/charliermarsh/ruff

:lock: [slack thread](https://nava.slack.com/archives/C3TRTAJ21/p1674850323551869)

chouinar:
## Ticket

Resolves #107

## Changes

Change our linter from flake8 to ruff

Python package updates

## Context for reviewers

[ruff](https://docs.astral.sh/ruff/) is a linter built in Rust that is both better performing and contains much more options than flake8 (including many of the Flake8 rules).

Ruff has a lot of [potential rules](https://docs.astral.sh/ruff/rules/) we could configure, but to keep things simple, I ported over the same rules that we already had configured. If we want to try out any of the other rules, we can follow up with that. For context, enabling all rules causes ~1300 errors, although many of the common ones I see are not issues I think we would want to actually consider (complaining about using `assert` in test functions)

Ruff can also serve as a formatter of code, potentially replacing black + isort, but I'll explore using that later as the change is likely much bigger as it would cause a lot of little formatting differences

## Testing
Was able to run ruff via `make lint-ruff` and only encountered a few small issues that I fixed.

We also switched to using this approach a few months ago in the simpler grants codebase and haven't run into any issues. https://github.com/HHS/simpler-grants-gov/pull/1626


docs/app/formatting-and-linting.md
```diff
@@ -17,7 +17,7 @@ When we run migrations via alembic, we autorun the formatters on the generated f
 Run `make lint` to run all of the linters. It's recommended you run the formatters first as they fix several linting issues automatically.
 
 ### Flake
```
KevinJBoyer:
```suggestion
### Ruff
```

docs/app/formatting-and-linting.md
```diff
@@ -17,7 +17,7 @@ When we run migrations via alembic, we autorun the formatters on the generated f
...
-[flake](https://flake8.pycqa.org/en/latest/) is used to validate the format of our Python code. Configuration options can be found in [setup.cfg](/app/setup.cfg).
+[ruff](https://flake8.pycqa.org/en/latest/) is used to enforce a set of best practices for our Python code. Configuration options can be found in [pyproject.toml - tool.ruff](/app/pyproject.toml).
```
KevinJBoyer, line 20:
```suggestion
[ruff](https://docs.astral.sh/ruff/) is used to enforce a set of best practices for our Python code. Configuration options can be found in [pyproject.toml - tool.ruff](/app/pyproject.toml).
```

docs/app/formatting-and-linting.md
```diff
@@ -17,7 +17,7 @@ When we run migrations via alembic, we autorun the formatters on the generated f
...
 
 We use two flake extensions:
```
KevinJBoyer: (And I think we should delete lines 22 - 24 here)

KevinJBoyer (approved):
LGTM, though I didn't review openapi.generated.yml since I'm not sure that's intentionally part of this PR?

chouinar:
> LGTM, though I didn't review openapi.generated.yml since I'm not sure that's intentionally part of this PR?

I upgraded the versions of the packages, looks like something in one of the libraries that generate the spec got changed, although I'm not sure of what specifically. Those were about a year out of date, and I remember this changing a bit when I upgraded the simpler grants repo. Shouldn't be an issue?

KevinJBoyer:
@chouinar one other note (I just noticed it while looking at my own PR for moving the docker compose files) -- can you update the docs/app/README.md to remove `setup.cfg` from there?

PR #235, opened 2024-08-26
Pin Python minor version and document upgrade details

chouinar:
## Changes

Pin the Python version and document details on how to upgrade

## Context for reviewers

Python releases new minor versions (3.12, 3.13, etc.) every year in October. It usually takes a few weeks for all of our dependencies and tooling to also be upgraded to the latest version, causing our builds to break. There isn't much we can do except wait a few weeks and then do the upgrade (assuming no breaking changes in features we use).

However, we had some of our dependencies pinned to the major version (Python 3) so it has broken the past few years until it started working again when the dependencies got fixed. This is just getting ahead of that and making sure the upgrade to Python 3.13 doesn't cause any problems.

## Testing
This change is largely documentation as the Python version used in the dockerfile + pyproject.toml already would have resolved to Python 3.12, this just makes it so it won't auto-upgrade to 3.13 when that releases in October.

### DIFF
--- a/docs/app/README.md
+++ b/docs/app/README.md
@@ -153,59 +153,29 @@ The API can be run in debug mode that allows for remote attach debugging (curren
 }
 ```
 
-# Upgrading Python
+## Upgrading Python
 Python does [yearly releases](https://devguide.python.org/versions/) for their minor versions (eg. 3.12 -> 3.13). They do not
 use semvar versioning, and their [minor releases](https://devguide.python.org/developer-workflow/development-cycle/#devcycle) contain
 breaking changes.
 
-We have this environment configured to use a specific minor version of python just in case anything would break when the yearly release does occur.
+Pin to a specific minor version of python just in case anything would break when the yearly release does occur.
+Only pin the minor versions of Python (eg. 3.12), and not the patch versions (eg. 3.12.1) as those contain bug and security fixes.
 
-We recommend as a starting point, only pin the minor versions of Python (eg. 3.12), and not the patch versions (eg. 3.12.1) as those contain bug and security fixes.
-However, your project may have dependencies, or tools that rely on very specific patch versions of Python, so adjust your approach as necessary.
+Along with any version upgrades, remember to:
+- Test the system functionality before deploying to production
+- Review the [changelog](https://docs.python.org/3/whatsnew/changelog.html)
+for any breaking changes of features you may use
 
-Along with any version upgrades, you should follow all best practices for upgrading software and test the functionality
-of your system before deploying to production. It is also recommended to review the [changelog](https://docs.python.org/3/whatsnew/changelog.html)
-for any breaking changes of features you may use.
-
-## Local Development
-If you run any of the python code outside of the Docker container, first upgrade your version
-of Python locally. The exact way you upgrade will depend on how you originally installed Python,
-but one strongly recommended approach is to use [pyenv](https://github.com/pyenv/pyenv) which allows
-you to install and swap between different python versions locally.
-
-If you run commands with Poetry, you'll also need to [configure](https://python-poetry.org/docs/managing-environments/#switching-between-environments)
-it to use the approach python version.
-
-## Dockerfile
-The dockerfile that our API is built from specifies the Python version as the first step.
-To upgrade this, simply change the version to the version you would like to use.
-
-For example, to upgrade from 3.12 to 3.13, change the following line:
-```dockerfile
-FROM python:3.12-slim AS base
-```
-to
-```dockerfile
-FROM python:3.13-slim AS base
-```
-
-## Poetry
-Adjust the pyproject.toml file line that specifies the version of Python used
-to your new version. For example, to upgrade from 3.12 to 3.13, change it from:
-
-```toml
-[tool.poetry.dependencies]
-python = "~3.12"
-```
-
-to
-```toml
-[tool.poetry.dependencies]
-python = "~3.13"
-```
-
-You will then need to run `poetry lock --no-update` to make the lock file reflect this change.
-
-## Misc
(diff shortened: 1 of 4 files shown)


docs/app/README.md
```diff
@@ -152,3 +152,60 @@ The API can be run in debug mode that allows for remote attach debugging (curren
     ]
 }
 ```
+
+# Upgrading Python
+Python does [yearly releases](https://devguide.python.org/versions/) for their minor versions (eg. 3.12 -> 3.13). They do not
+use semvar versioning, and their [minor releases](https://devguide.python.org/developer-workflow/development-cycle/#devcycle) contain
+breaking changes.
+
+We have this environment configured to use a specific minor version of python just in case anything would break when the yearly release does occur.
```
lorenyu:
```suggestion
Pin to a specific minor version of python just in case anything would break when the yearly release does occur.
```

docs/app/README.md
```diff
@@ -152,3 +152,60 @@ The API can be run in debug mode that allows for remote attach debugging (curren
...
+
+We recommend as a starting point, only pin the minor versions of Python (eg. 3.12), and not the patch versions (eg. 3.12.1) as those contain bug and security fixes.
```
lorenyu:
```suggestion
Only pin the minor versions of Python (eg. 3.12), and not the patch versions (eg. 3.12.1) as those contain bug and security fixes.
```

docs/app/README.md
```diff
@@ -152,3 +152,60 @@ The API can be run in debug mode that allows for remote attach debugging (curren
...
 }
 ```
+
+# Upgrading Python
+Python does [yearly releases](https://devguide.python.org/versions/) for their minor versions (eg. 3.12 -> 3.13). They do not
+use semvar versioning, and their [minor releases](https://devguide.python.org/developer-workflow/development-cycle/#devcycle) contain
+breaking changes.
+
+We have this environment configured to use a specific minor version of python just in case anything would break when the yearly release does occur.
+
```
lorenyu:
```suggestion
```

docs/app/README.md
```diff
@@ -152,3 +152,60 @@ The API can be run in debug mode that allows for remote attach debugging (curren
...
+We recommend as a starting point, only pin the minor versions of Python (eg. 3.12), and not the patch versions (eg. 3.12.1) as those contain bug and security fixes.
+However, your project may have dependencies, or tools that rely on very specific patch versions of Python, so adjust your approach as necessary.
```
lorenyu: To simplify I think we can remove this line. Unless you think this is pretty common

```suggestion
```

docs/app/README.md
```diff
@@ -152,3 +152,60 @@ The API can be run in debug mode that allows for remote attach debugging (curren
...
+
+Along with any version upgrades, you should follow all best practices for upgrading software and test the functionality
+of your system before deploying to production. It is also recommended to review the [changelog](https://docs.python.org/3/whatsnew/changelog.html)
+for any breaking changes of features you may use.
```
lorenyu:
```suggestion
Along with any version upgrades, remember to:
- Test the system functionality before deploying to production
- Review the [changelog](https://docs.python.org/3/whatsnew/changelog.html)
for any breaking changes of features you may use
```

docs/app/README.md
```diff
@@ -152,3 +152,60 @@ The API can be run in debug mode that allows for remote attach debugging (curren
...
+
+## Local Development
+If you run any of the python code outside of the Docker container, first upgrade your version
+of Python locally. The exact way you upgrade will depend on how you originally installed Python,
+but one strongly recommended approach is to use [pyenv](https://github.com/pyenv/pyenv) which allows
+you to install and swap between different python versions locally.
+
+If you run commands with Poetry, you'll also need to [configure](https://python-poetry.org/docs/managing-environments/#switching-between-environments)
+it to use the approach python version.
```
lorenyu: could we just tell the developer to "update their local version of python and link to "getting-started.md" for deeper instructions"?

docs/app/README.md
```diff
@@ -152,3 +152,60 @@ The API can be run in debug mode that allows for remote attach debugging (curren
     ]
 }
 ```
+
+# Upgrading Python
```
lorenyu: shift all the headings down a level e.g. this should be h2 `##` instead of `#`. (there should only be one h1 per document and that should be the title which is at the top)

docs/app/README.md
```diff
@@ -152,3 +152,60 @@ The API can be run in debug mode that allows for remote attach debugging (curren
...
+```toml
+[tool.poetry.dependencies]
+python = "~3.13"
+```
+
+You will then need to run `poetry lock --no-update` to make the lock file reflect this change.
+
+## Misc
+Some tools, include pyenv, reference the [.python-version](/app/.python-version) file in order to determine which
+version of python should be used. Update this to the approach version as well.
```
lorenyu: Could we make this whole section more concise like:

```
To upgrade the Python version, make changes in the following places:
1. Local Python version (see more about managing local Python versions in [getting started](/docs/app/getting-started.md))
2. [Dockerfile](/app/Dockerfile)
    search for the line `FROM python:3.12-slim as base`
3. [pyproject.toml](/app/pyproject.toml)
    search for the line
    ```
    [tool.poetry.dependencies]
    python = "~3.12"
    ```
4. .python-version
```

I'm trying to think of ways to improve skimmability since this doc is pretty long and I think having a bunch of subheadings might make it harder to skim

lorenyu (approved):
looks good, left some suggestions

docs/app/README.md
chouinar: I can minimize it a bit, but there are some important steps like in the pyproject file needing to run the Poetry command to fix the lock file.

PR #237, opened 2024-10-28
Upgrade dependencies + upgrade Python version

chouinar:
## Changes
Upgraded python version from 3.12 to 3.13

Upgraded many package versions of other dependencies

## Context for reviewers
This should be uneventful, the upgrade from Python 3.12 -> 3.13 had very little that would be breaking (many of the major changes need to be enabled explicitly).

The package versions are just catching up on updates - we've been doing a better job of this in the simpler grants repo. There are several vulnerabilities this should fix. https://github.com/navapbc/template-application-flask/security/dependabot


coilysiren (approved):
💹

PR #239, opened 2024-12-10
Update packages & setup docker-compose to allow override.env file

chouinar:
## Changes
Updated Python packages

Add support for an override.env file

## Context for reviewers
These python package updates mirror similar ones we've done already on Simpler Grants.gov, so not concerned with any of them.

The override.env file uses something added to Docker in January 2024: https://docs.docker.com/reference/compose-file/services/#required - in short, you can specify multiple .env files and have some be optional. This way we can always say to look for override.env, but not require it, and allow a user to write whatever they want to it without checking it in.

## Testing
Locally I was able to create override.env, it not be added to git, and the overrides worked (changed `LOG_FORMAT` to `json` and back).

### DIFF
--- a/app/.gitignore
+++ b/app/.gitignore
@@ -19,6 +19,7 @@ coverage.*
 # Environment variables
 .env
 .envrc
+override.env
 
 # mypy
 .mypy_cache
--- a/docs/app/README.md
+++ b/docs/app/README.md
@@ -97,7 +97,7 @@ direnv: export +API_AUTH_TOKEN +AWS_ACCESS_KEY_ID +AWS_DEFAULT_REGION +AWS_SECRE
 Most configuration options are managed by environment variables.
 
 Environment variables for local development are stored in the [local.env](/app/local.env) file. This file is automatically loaded when running. If running within Docker, this file is specified as an `env_file` in the [docker-compose](/app/docker-compose.yml) file, and loaded [by a script](/app/src/util/local.py) automatically when running unit tests (see running natively above for other cases).
-To override environment variables for locally, create an `override.env` file in the same directory as local.env. The [docker-compose file](https://docs.docker.com/reference/compose-file/services/#required) optionally checks this file as well.
+To override environment variables for locally, create an `override.env` file in the same directory as local.env. The docker-compose file [optionally checks this file](https://docs.docker.com/reference/compose-file/services/#required) as well.
 
 Any environment variables specified directly in the [docker-compose](/app/docker-compose.yml) file will take precedent over those specified in the [local.env](/app/local.env) file.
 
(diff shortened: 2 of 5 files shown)


app/.gitignore
```diff
@@ -19,6 +19,7 @@ coverage.*
 # Environment variables
 .env
 .envrc
+override.env
```
lorenyu, line 22: would this work
```suggestion
*.env
.envrc
```

docs/app/README.md
```diff
@@ -97,6 +97,7 @@ direnv: export +API_AUTH_TOKEN +AWS_ACCESS_KEY_ID +AWS_DEFAULT_REGION +AWS_SECRE
 Most configuration options are managed by environment variables.
 
 Environment variables for local development are stored in the [local.env](/app/local.env) file. This file is automatically loaded when running. If running within Docker, this file is specified as an `env_file` in the [docker-compose](/app/docker-compose.yml) file, and loaded [by a script](/app/src/util/local.py) automatically when running unit tests (see running natively above for other cases).
+If you would like to override environment variables for your own local development, create an `override.env` file in the same directory as local.env. The [docker-compose file](https://docs.docker.com/reference/compose-file/services/#required) optionally checks this file as well.
```
lorenyu: nit
```suggestion
To override environment variables locally, create an `override.env` file in the same directory as local.env. The [docker-compose file](https://docs.docker.com/reference/compose-file/services/#required) optionally checks this file as well.
```

lorenyu (approved):

app/.gitignore
chouinar: If I do `*.env` we'd exclude local.env which we do want to check-in.

KevinJBoyer (approved):

docs/app/README.md
```diff
@@ -97,6 +97,7 @@ direnv: export +API_AUTH_TOKEN +AWS_ACCESS_KEY_ID +AWS_DEFAULT_REGION +AWS_SECRE
 Most configuration options are managed by environment variables.
 
 Environment variables for local development are stored in the [local.env](/app/local.env) file. This file is automatically loaded when running. If running within Docker, this file is specified as an `env_file` in the [docker-compose](/app/docker-compose.yml) file, and loaded [by a script](/app/src/util/local.py) automatically when running unit tests (see running natively above for other cases).
+To override environment variables for locally, create an `override.env` file in the same directory as local.env. The [docker-compose file](https://docs.docker.com/reference/compose-file/services/#required) optionally checks this file as well.
```
KevinJBoyer:
```suggestion
To override environment variables for locally, create an `override.env` file in the same directory as local.env. The docker-compose file [optionally checks this file](https://docs.docker.com/reference/compose-file/services/#required) as well.
```

a nit to move the hyperlink to the part of the text that it explains

KevinJBoyer: (this might not actually be clearer so feel free to ignore)

PR #243, opened 2025-04-14
add to test coverage

lisac:
## Ticket

n/a

## Changes

Copies over unit tests discovered in other repositories, where the subjects of the tests are functions made available by the template flask application.

## Testing

https://github.com/navapbc/platform-test-flask/pull/9


lorenyu:
@lisac thanks! for testing, could you make these same changes in a PR in https://github.com/navapbc/platform-test-flask so that we can make sure these changes passes CI?

lisac:
hi @lorenyu ! my apologies, this is my first time working in this repo.
For testing, I created https://github.com/navapbc/platform-test-flask/pull/9 and was able to get the CI job for app test to succeed (https://github.com/navapbc/platform-test-flask/actions/runs/14456152418/job/40539734786?pr=9). Is there another piece I need to account for?

template/{{app_name}}/tests/src/logging/test_audit.py
```diff
@@ -146,6 +146,25 @@ def test_audit_hook(
         assert_record_match(record, expected_record)
 
 
+def test_get_addr_info(init_audit_hook, caplog: pytest.LogCaptureFixture):
+    socket.getaddrinfo("www.python.org", 80)
+    print(caplog.records)
```
chouinar:
```suggestion

```

template/{{app_name}}/tests/src/logging/test_audit.py
```diff
@@ -146,6 +146,24 @@ def test_audit_hook(
         assert_record_match(record, expected_record)
 
 
+def test_get_addr_info(init_audit_hook, caplog: pytest.LogCaptureFixture):
```
lorenyu: this test is redundant with `test_audit_hook`, there's already a use case for socket.getaddrinfo (see https://github.com/navapbc/template-application-flask/blob/12c5d00e09298597b7767c78ab0686f749af2536/template/%7B%7Bapp_name%7D%7D/tests/src/logging/test_audit.py#L96-L101)

template/{{app_name}}/tests/src/test_app_config.py
```diff
@@ -0,0 +1,11 @@
+import os
+from unittest import mock
+
+from src.app_config import AppConfig
+
+
+@mock.patch.dict(os.environ, {"host": "192.123.123.123", "port": "5190"})
+def test_app_config_sets_values_from_environment():
```
lorenyu: I'm on the fence about including this test and lean somewhat against including this test because it seems like it's kind of [retesting pydantic's BaseSettings class](https://docs.pydantic.dev/latest/api/pydantic_settings/). What sort of regression/bug is this test attempting to protect against?

template/{{app_name}}/tests/src/util/test_string_utils.py
```diff
@@ -0,0 +1,14 @@
+from src.util.string_utils import join_list
+
+
+def test_join_list():
+    assert join_list(None) == ""
```
lorenyu, line 5: out of scope of this PR but why do we even have this function

lorenyu (approved):
test_audit test is redundant and I don't think we should include the test_app_config test, everything else looks great, thanks!

template/{{app_name}}/tests/src/test_app_config.py
lisac: frankly, i was reacting to the test coverage % on that file.  I agree with your point that it's retesting pydantic's implementation. have removed!

PR #245, opened 2025-08-05
Update docker-compose.yml.jinja

KevinJBoyer:
## Ticket

n/a

## Changes
Make the local container for development use the port specified in the platform-cli tool

## Testing
Install the flask template using the CLI, use the default suggested port of 3000. The docs will update to say that you can access it after running `make init start` at localhost:3000, but you'll need to update the docker-compose as suggested in this PR first.


doshitan:
Ah yea, I included a similar change in https://github.com/navapbc/template-application-flask/commit/ad745688d9f66e8502fe5298e875cb1ab9621930 as I was updating other things.

PR #249, opened 2025-08-25
Configure isort to follow black formatting conventions

KevinJBoyer:
## Ticket

n/a

## Changes
 - Configure isort to follow black formatting conventions

## Context for reviewers
 - Without this change there are situations where isort's formatting conflicts with blacks, leading `make format-check` to fail even after `make-format`

## Testing
You can see an example file that will fail without this change at this PR: https://github.com/navapbc/labs-referral-pilot/pull/11

### DIFF
--- a/template/{{app_name}}/pyproject.toml
+++ b/template/{{app_name}}/pyproject.toml
@@ -55,11 +55,6 @@ line-length = 100
 
 [tool.isort]
 profile = "black"
-multi_line_output = 3
-include_trailing_comma = true
-force_grid_wrap = 0
-use_parentheses = true
-line_length = 100
 
 [tool.ruff]
 line-length = 100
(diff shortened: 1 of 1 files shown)


template/{{app_name}}/pyproject.toml
```diff
@@ -54,6 +54,7 @@ db-migrate-down-all = "src.db.migrations.run:downall"
 line-length = 100
 
 [tool.isort]
+profile = "black"
 multi_line_output = 3
 include_trailing_comma = true
 force_grid_wrap = 0
```
KevinJBoyer: Copilot suggests that these lines are redundant with `profile = "black"` but 🤷

doshitan: Yea, looks like profiles were introduced in isort v5, which we require, so safe to drop the settings listed upstream for "black": https://github.com/PyCQA/isort/blob/5.13.2/isort/profiles.py#L4-L10

doshitan (approved):
Looks good. I'd rather switch to ruff formatting instead of black+isort, but good to improve the existing stuff where easy.

Can drop the redundant settings before merge.

PR #253, opened 2025-11-07
Always use "app" in dockle accept file path

### LINKED ISSUE #252: Dockle accept files points at wrong path
When an app name that isn't "app" is specified for the template, the dockle accept files in `template/{{app_name}}/.dockleconfig.jinja` are incorrect. That settings.py file is always at `app/.venv/lib/python3.13/site-packages/apiflask/settings.py` regardless of the app name because the working directory in the Dockerfile is always "app".

gsf:
## Ticket

Resolves #252

## Changes

Fixed path for settings.py in DOCKLE_ACCEPT_FILES.

## Context for reviewers

Can be tested locally by exporting DOCKLE_ACCEPT_FILES before running dockle. Without this fix, the settings.py file in the app causes that check to fail. See the Dockerfile for where "app" is hardcoded in the path: https://github.com/navapbc/template-application-flask/blob/5d4d40aaa45d217d19e65814ff5f18cdff4b04fb/template/%7B%7Bapp_name%7D%7D/Dockerfile#L77

## Testing

Tested in a generated app.


doshitan (approved):
Thank you! We should also rename `template/{{app_name}}/.dockleconfig.jinja` to `template/{{app_name}}/.dockleconfig` here since it doesn't need templated now.

